PR #75, opened 2017-12-08
Combine the expanded output and auto expand options in config.

amjith:
This combines two separate options in the config file into a single one. 

Be aware this does change the default behavior of the output. The expanded mode is now set to auto, meaning if the table is too wide to fit the screen then the output will be shown using the vertical mode instead of the tabular mode.

### DIFF
--- a/mssqlcli/main.py
+++ b/mssqlcli/main.py
@@ -135,9 +135,8 @@ class MssqlCli(object):
         self.multi_line = c['main'].as_bool('multi_line')
         self.multiline_mode = c['main'].get('multi_line_mode', 'tsql')
         self.vi_mode = c['main'].as_bool('vi')
-        self.auto_expand = auto_vertical_output or c['main'].as_bool(
-            'auto_expand')
-        self.expanded_output = c['main'].as_bool('expand')
+        self.auto_expand = auto_vertical_output or c['main']['expand'] == 'auto'
+        self.expanded_output = c['main']['expand'] == 'always'
         if row_limit is not None:
             self.row_limit = row_limit
         else:
--- a/mssqlcli/mssqlclirc
+++ b/mssqlcli/mssqlclirc
@@ -18,11 +18,11 @@ multi_line = False
 # a command.
 multi_line_mode = tsql
 
-# Enables expand mode.
-expand = False
-
-# Enables auto expand mode.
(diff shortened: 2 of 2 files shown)


MrMeemus:
@amjith @pensivebrian, @yualan had to delete the config file in order to pick up this change. What would be the best way to communicate this to the users? or should we detect if auto_expand exists and then delete it and create the config file? <- not sure if this is a good approach

pensivebrian:
Good question. IMHO, migrating settings should just happen, so we should programmatically patch these situations on first run after upgrade. However, this change went in before public preview, so I didn't worry about it. Going forward, we should handle change like these more gracefully. With PR https://github.com/dbcli/mssql-cli/pull/137, we have a new setting, which ideally would be added to all configs.

MrMeemus:
@pensivebrian I definitely with migrating settings. The odds of user's getting new settings are slim if they don't fiddle with it often. I can add support to detect if there is a diff for settings and either add it to the file.

MrMeemus:
@pensivebrian , what options did you use for configuring less? using the default -SRXF jumbles up the output.

pensivebrian:
I didn't set any less configuration options - just dropped it in the path. IIRC, things mostly worked, but there were some issues, but I didn't explore deeply... just a quick run through.

CODE-REaD:
@amjith, thank you for helping make mssql-cli a most welcome tool.  Above you write,
> We do that by setting the LESS environment variable and we only do that if the user hasn't defined that variable.

I think you meant to say the **_PAGER_** environment variable rather than **_LESS_**.  At least this is how mssql-cli 1.0.0 operates for me under Ubuntu 16.04.4 LTS.

In my case I needed to set my PAGER environment variable in order to disable horizontal scrolling so that I could cut and paste output (when less scrolls horizontally, only the section of output present on the screen is copied when highlighted).

I was led to this discovery when I found my LESS environment variable (which may be used to configure options for the `less` command) was already set to `-i` and mssql-cli was ignoring this setting.  Reading mssql-cli's configuration file, I found,
> Default pager.
By default 'PAGER' environment variable is used
pager = less -SRXF

And by entering the command `export PAGER="less -i"` on my command line before running `mssql-cli` I was able to disable horizontal scrolling.

Issue #118, opened 2017-12-18
Allow for full SQL Server connection properties


rahulballal:
At the moment thee is no way to specify connection properties like `ApplicationIntent`. It would be awesome to see this in upcoming releases.

https://blogs.msdn.microsoft.com/alwaysonpro/2013/08/02/connect-to-sql-server-using-application-intent-read-only/

abhisheksinha89:
Thanks, we plan on doing this 👍

abhisheksinha89:
The fix went in as part of the above PR. . It will be resolved in our next release.

abhisheksinha89:
Please update mssql-cli. It should be fixed in v0.9.0 and later. 
pip install --upgrade mssql-cli

rahulballal:
great work @abhisheksinha89 !! Thanks

PR #126, opened 2017-12-21
Abhi/connection properties

abhisheksinha89:
PR for
https://github.com/dbcli/mssql-cli/issues/118

### DIFF
--- a/mssqlcli/jsonrpc/contracts/connectionservice.py
+++ b/mssqlcli/jsonrpc/contracts/connectionservice.py
@@ -85,12 +85,26 @@ class ConnectionDetails(object):
     """
 
     def __init__(self, parameters):
+        # Required params
         self.ServerName = parameters[u'ServerName']
         self.DatabaseName = parameters[u'DatabaseName']
         self.UserName = parameters[u'UserName']
         self.Password = parameters[u'Password']
         self.AuthenticationType = parameters[u'AuthenticationType']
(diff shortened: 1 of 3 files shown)


abhisheksinha89:
I'm currently brainstorming the best ideas to write tests for these connection params that does not involve setting up an actual live server with certs for encryption, etc.
Any suggestions welcome.

abhisheksinha89:
I've only done manual testing, will not merge till I have unit tests for them

---
In reply to: [353253063](https://github.com/dbcli/mssql-cli/pull/126#issuecomment-353253063) [](ancestors = 353253063)

MrMeemus:
@abhisheksinha89 , You could perhaps move the click decorators into their own module to make it modular for unit testing. Then you can use the CLI runner which is part of click.testing module to run through the options to ensure they are populated. That should verify the input aspect, then you can submit those options to the connection service contract and verify the format of the output.

For reference for click testing:
http://click.pocoo.org/5/testing/

Let me know if you thought of this already and found a issue with it.

abhisheksinha89:
Talked with Brian, manual testing should be good for now as we are focused more on unblocking customers. I'll add an issue to write unit tests for command line params

---
In reply to: [353370093](https://github.com/dbcli/mssql-cli/pull/126#issuecomment-353370093) [](ancestors = 353370093)

Issue #129, opened 2017-12-27
Stored procedure reporting multiple result-sets are not displayed in mssql-cli results


sethusrinivasan:
1) Create a database test1
2) Create  following stored proc
CREATE OR ALTER PROC sp_test_multiple_results
AS
BEGIN
SELECT TOP 10 name,object_id from sys.objects
SELECT TOP 10 name,object_id from sys.columns
END

3) run mssql-cli and execute this proc

**Outcome:**
Only first result set is displayed
mssql-cli
![image](https://user-images.githubusercontent.com/3907049/34367290-9bd2ad76-ea5d-11e7-8f41-d0f779f6543e.png)

**Expected:**
All resultsets to be displayed to user and should be consistent with other tools like SSMS / sqlcmd

SSMS:
![image](https://user-images.githubusercontent.com/3907049/34367302-c9e3ab02-ea5d-11e7-8198-9d73554d722a.png)

sqlcmd:
![image](https://user-images.githubusercontent.com/3907049/34367306-fbc25538-ea5d-11e7-8968-7ee2da98b163.png)

abhisheksinha89:
Will look into this

abhisheksinha89:
Found the root cause of this issue, working on a fix

abhisheksinha89:
@pensivebrian @sethusrinivasan : Above PR to fix this issue

abhisheksinha89:
This should be fixed in 0.9.0 and later

PR #130, opened 2018-01-03
Support multiple result sets

abhisheksinha89:
https://github.com/dbcli/mssql-cli/issues/129

### DIFF
--- a/mssqlcli/mssqlcliclient.py
+++ b/mssqlcli/mssqlcliclient.py
@@ -215,55 +215,63 @@ class MssqlCliClient(object):
         """ Returns a list of schema names"""
         query = mssqlqueries.get_schemas()
         logger.info(u'Schemas query: {0}'.format(query))
-        return [x[0] for x in self.execute_single_batch_query(query)[0]]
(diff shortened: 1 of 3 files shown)


pensivebrian:
>     def test_mssqlcliclient_multiple_statement(self):

Can we add two more tests:

1) two statements, the first one valid. Expected: the first statement executes, and then we see an error:

select 1; select 'I' from teapot

2) two statements, the first one invalid. Expected: the first statement fails and the second does not execute.

select 'me' from teapot; select 1; 


---
Refers to: tests/test_mssqlcliclient.py:150 in ef873d5. [](commit_id = ef873d5d9201cf1066b317d3146dbd3fcd638a62, deletion_comment = False)

pensivebrian:
> class MssqlCliClientTests(unittest.TestCase):

Can we add a test that runs a stored proc that returns multiple result sets?

---
Refers to: tests/test_mssqlcliclient.py:21 in ef873d5. [](commit_id = ef873d5d9201cf1066b317d3146dbd3fcd638a62, deletion_comment = False)

PR #137, opened 2018-01-10
Add prompt format as config and CLI option.

j-bennet:
Configurable prompt for mssql-cli. I set the default to `"\\d> "` to mimic the current behavior (even though pgcli and mycli have longer default prompts, e.g. `"\\u@\\h:\\d> "`.

### DIFF
--- a/mssqlcli/main.py
+++ b/mssqlcli/main.py
@@ -98,6 +98,7 @@ Microsoft Privacy statement: https://privacy.microsoft.com/en-us/privacystatemen
 class MssqlCli(object):
 
     max_len_prompt = 30
+    default_prompt = '\\d> '
 
     def set_default_pager(self, config):
         configured_pager = config['main'].get('pager')
@@ -125,7 +126,8 @@ class MssqlCli(object):
                  mssqlclirc_file=None, row_limit=None,
                  single_connection=False, less_chatty=None,
                  auto_vertical_output=False, sql_tools_client=None,
-                 integrated_auth=False, enable_sqltoolsservice_logging=False):
+                 integrated_auth=False, enable_sqltoolsservice_logging=False,
+                 prompt=None):
 
         self.force_passwd_prompt = force_passwd_prompt
 
@@ -143,6 +145,7 @@ class MssqlCli(object):
         self.vi_mode = c['main'].as_bool('vi')
         self.auto_expand = auto_vertical_output or c['main']['expand'] == 'auto'
         self.expanded_output = c['main']['expand'] == 'always'
+        self.prompt_format = prompt or c['main'].get('prompt', self.default_prompt)
         if row_limit is not None:
             self.row_limit = row_limit
         else:
@@ -435,7 +438,7 @@ class MssqlCli(object):
             set_vi_mode_enabled=set_vi_mode)
 
         def prompt_tokens(_):
-            prompt = self.get_prompt()
+            prompt = self.get_prompt(self.prompt_format)
             return [(Token.Prompt, prompt)]
 
         def get_continuation_tokens(cli, width):
@@ -648,9 +651,14 @@ class MssqlCli(object):
             return self.completer.get_completions(
                 Document(text=text, cursor_position=cursor_positition), None)
 
-    def get_prompt(self):
+    def get_prompt(self, string):
         # mssql-cli
-        string = self.mssqlcliclient_query_execution.database + u'>'
+        string = string.replace('\\t', self.now.strftime('%x %X'))
+        string = string.replace('\\u', self.mssqlcliclient_query_execution.user_name or '(none)')
+        string = string.replace('\\h', self.mssqlcliclient_query_execution.host or '(none)')
+        string = string.replace('\\d', self.mssqlcliclient_query_execution.database or '(none)')
+        string = string.replace('\\p', str(self.mssqlcliclient_query_execution.port) or '(none)')
+        string = string.replace('\\n', "\n")
         return string
 
     def get_last_query(self):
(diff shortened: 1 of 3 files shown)


mssqlcli/main.py
```diff
@@ -98,6 +98,7 @@
 class MssqlCli(object):
 
     max_len_prompt = 30
+    default_prompt = '\\d> '
```
pensivebrian: I don't think a trailing space is a common convention. Can we remove the space after the '>' for the default?

mssqlcli/main.py
```diff
@@ -698,9 +706,10 @@ def get_last_query(self):
 @click.option('--enable-sqltoolsservice-logging', is_flag=True,
               default=False,
               help='Enables diagnostic logging for the SqlToolsService.')
+@click.option('--prompt', help='Prompt format (Default: "\\d> ").')
```
pensivebrian: Can we reference the `default_prompt` constant above?

mssqlcli/mssqlcliclient.py
```diff
@@ -24,7 +24,13 @@ def __init__(self, sql_tools_client, server_name, user_name, password,
                  authentication_type=u'SqlLogin', database=u'master', owner_uri=None, multiple_active_result_sets=True,
                  encrypt=None, trust_server_certificate=None, connection_timeout=None, application_intent=None,
                  multi_subnet_failover=None, packet_size=None, **kwargs):
+
         self.server_name = server_name
+        if ',' in server_name:
+            self.host, self.port = self.server_name.split(',')
```
pensivebrian: since host and port are for prompt display purposes, can you rename them to prompt_host and prompt_port? This way, it's clear we're not using the values when connecting to the target database.

mssqlcli/mssqlclirc
```diff
@@ -103,6 +103,15 @@ row_limit = 1000
...
 less_chatty = False
 
+# mysql-cli prompt
+# \t - Current date and time
+# \u - Username
+# \h - Hostname of the server
+# \d - Database name
+# \p - Database port
+# \n - Newline
+prompt = '\d> '
```
pensivebrian: Same as above, let's we remove the space after the '>' for the default.

mssqlcli/main.py
```diff
@@ -648,9 +651,14 @@ def get_completions(self, text, cursor_positition):
             return self.completer.get_completions(
                 Document(text=text, cursor_position=cursor_positition), None)
 
-    def get_prompt(self):
+    def get_prompt(self, string):
         # mssql-cli
-        string = self.mssqlcliclient_query_execution.database + u'>'
+        string = string.replace('\\t', self.now.strftime('%x %X'))
+        string = string.replace('\\u', self.mssqlcliclient_query_execution.user_name or '(none)')
```
pensivebrian: If the `-E` options is specified, we use windows/kerberos, which defaults to the current user. This is really common for Windows environments. Can we detect if -E was passed, and get the current user via `os.getlogin()`?

pensivebrian: If you don't have a Windows environment to test this, feel free to skip this, and someone with a Windows machine can fix this in another PR.

mssqlcli/mssqlcliclient.py
```diff
@@ -24,7 +24,13 @@ def __init__(self, sql_tools_client, server_name, user_name, password,
...
+        else:
+            self.host = server_name
+            self.port = 1433
```
pensivebrian: For practical purposes, 1433 is correct. However, SQL Server supports other transports, named pipes and shared memory, which use different ports. Also, the dedicated admin connection uses 1434. However, we can ignore this and fix if we ever get any complaints. I doubt we will :)

mssqlcli/mssqlclirc
```diff
@@ -103,6 +103,15 @@ row_limit = 1000
 # Skip intro on startup and goodbye on exit
 less_chatty = False
 
+# mysql-cli prompt
```
pensivebrian: mssql-cli :)

mssqlcli/main.py
j-bennet: I think it is. Both mycli and pgcli use trailing ` ` in their prompt:

https://github.com/dbcli/pgcli/blob/327f15afc3fcbd556b87a5b568b2b6c983519bdd/pgcli/pgclirc#L123

https://github.com/dbcli/mycli/blob/d6ccd390a9b4b659a2bd35e8bf94172eb9c3874e/mycli/myclirc#L67

Without the ` `, once you start typing, it looks like the prompt and the query are glued together:

```
foo>select 1
```

not pretty!

```
foo> select 1
```

pretty!

mssqlcli/mssqlclirc
j-bennet: Yiiikes. :) Good catch.

j-bennet:
@pensivebrian I addressed all the concerns, except one - I'd still prefer to keep the trailing space in prompt.

Would you mind trying out the `os.getlogin()` change on windows? Indeed, I do not have access to a Windows machine.

Thank you!

mssqlcli/main.py
pensivebrian: Fair enough. I don't feel strongly either way, so a space it is.

pensivebrian:
@j-bennet I led you astray with the `os.getlogin()` suggestion. Turns out, this function doesn't exist on Windows. It appears `getpass.getuser()` from the [getpass ](https://docs.python.org/2.7/library/getpass.html) module is the proper way to get the current user for Windows and Linux.

If you push the getpass change, I'll give it another go on Windows. My apologies for the bad advice.

j-bennet:
No problem - changed to `getpass.getuser()`.

pensivebrian:
Thanks! Worked perfectly on Windows.

pensivebrian:
Paging @abhisheksinha89 and @MrMeemus - I plan on merging this tomorrow morning. If you can, please take a look. Thanks!

Issue #136, opened 2018-01-09
Missing "quotation mark" - mssql_cli stuck


Komport:
I am testing simple stuff with mssql_cli on Centos 7.0 and Python 2.7. When i am doing simple insert command as below in one point I forget to put closing quotation mark after some value. mssql_cli stuck at this point and only Ctrl+C can help.
```
DBSIZE>INSERT INTO test_table (some_value) values('Testing);
^CCancelling query...
Query cancelled.
DBSIZE>INSERT INTO test_table (some_value) values("Testing);
```

Komport:
After İ have waited a little longer and hit Ctrl+C I got below TraceBack

```
DBSIZE>INSERT INTO test_table (some_value) values("Testing);
";
^CCancelling query...
Traceback (most recent call last):
  File "/usr/lib64/python2.7/runpy.py", line 162, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib64/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/usr/lib/python2.7/site-packages/mssqlcli/main.py", line 867, in <module>
    cli()
  File "/usr/lib/python2.7/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/usr/lib/python2.7/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/lib/python2.7/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/mssqlcli/main.py", line 736, in cli
    mssqlcli.run_cli()
  File "/usr/lib/python2.7/site-packages/mssqlcli/main.py", line 413, in run_cli
    query = self.execute_command(document.text, query)
  File "/usr/lib/python2.7/site-packages/mssqlcli/main.py", line 342, in execute_command
    self.mssqlcliclient_query_execution)
  File "/usr/lib/python2.7/site-packages/mssqlcli/mssqlcliclient.py", line 304, in reset_connection_and_clients
    raise e
OSError: [Errno 3] No such process

```

MrMeemus:
Thanks for reporting, it could be a mishandled error from sqltoolsservice leaving the process hanging. Will take a look.

[...]

PR #140, opened 2018-01-11
Fix for malformed query leading to CLI hanging.

MrMeemus:
A query execute string request would not mark itself as finished when receiving a query execute error response event, leaving mssqlcliclient infinitely looping for a response.

If a error is returned during execute_single_batch_query, immediately return rather.

Fix for #136


mssqlcli/mssqlcliclient.py
```diff
@@ -139,18 +139,14 @@ def execute_single_batch_query(self, query):
         while not query_request.completed():
             query_response = query_request.get_response()
             if query_response:
-                if isinstance(query_response, queryservice.QueryExecuteErrorResponseEvent):
-                    yield self.tabular_results_generator(column_info=None, result_rows=None,
-                                                          query=query, message=query_response.error_message,
-                                                          is_error=True)
-                elif isinstance(query_response, queryservice.QueryMessageEvent):
```
abhisheksinha89: We are we deleting this? We need it to hamdle Error response events

abhisheksinha89: Ahh ok I see now what happened. Thanks

---
In reply to: [161077826](https://github.com/dbcli/mssql-cli/pull/140#discussion_r161077826) [](ancestors = 161077826)

PR #145, opened 2018-01-24
Ron/build fix

MrMeemus:
When the fix for issue #136  was submitted, it used syntax that is not supported on Python <= 3.3. The error in particular is using a return statement within a function that returns a generator. The correct fix is to first yield the message we want to return and then proceed to return from the function call.

The test was also updated because if a query fails, the request finishes on a query error event. Finishing on a error event does not yield a query complete event.

Tests have been rerun and lab runs looked clean.


pensivebrian:
Looks good, but we don't support Python 3.3. Fix still apply?

MrMeemus:
@pensivebrian, The fix still applies to anything less than python 3.3.

MrMeemus:
Windows CI build fails due to connection error with the test server. Will merge PR with confidence from successful runs on the other platforms.

Issue #111, opened 2017-12-14
Different display of cli in PowerShell and Command Prompt


bonskijr:
mssql-cli displays differently when being used inPowershell console. In command prompt it displays as it should but in Powershell it display filler "rectangle" for spaces.

Poweshell

![image](https://user-images.githubusercontent.com/60311/33982923-54263c96-e0ed-11e7-9a51-df29c41e73d7.png)

Command Prompt

![image](https://user-images.githubusercontent.com/60311/33982954-61518556-e0ed-11e7-9802-d5d5ad1d48b9.png)

mssql-cli v 0.7 .0
OS: Windows 10.0.15063
PowerShell : 5.1.15063.726

abhisheksinha89:
Hi, I am not able to repro this on Win10 Powershell 5.1 
Can you tell me what codepage you are using for the powershell terminal?

bonskijr:
The original codepage was 437 (US)
![image](https://user-images.githubusercontent.com/60311/34187364-fa882d2e-e56b-11e7-941b-42cd056fdfd6.png)

Then after running the cli changed to 65001 (UTF-8)

![image](https://user-images.githubusercontent.com/60311/34187350-e5072fb8-e56b-11e7-931c-346c7e99836d.png)

But as you can see both the issues, at least for me

abhisheksinha89:
Thanks I'll need to investigate this. I can't seem to repro this on 2 of our machines using version 0.8.0.
I noticed that you still have version 0.7.0.
Try doing a pip install --upgrade mssql-cli 

Eventhough I think we didn't make any major changes just try upgrading and see if the problem persists. I'll try to get a repro so I can investigate this on my end.

bonskijr:
Hi @abhisheksinha89 ,

I tried upgrading but issue still persist, but thanks for looking into this.

![image](https://user-images.githubusercontent.com/60311/34187771-fd956aac-e56d-11e7-992d-bf88e5de4faa.png)

snewfie:
I having the issue also and using 65001(UTF-8) code page is there a fix in sight.
using windows 10 and powershell shell also occurs on cmd shell.

![image](https://user-images.githubusercontent.com/4139281/34566152-90c98cea-f12b-11e7-889f-7a3a25cef427.png)
![image](https://user-images.githubusercontent.com/4139281/34566194-b5df94de-f12b-11e7-9861-adee4f8ff629.png)

But if I use the cmd shell in legacy mode works fine 

[...]

PR #153, opened 2018-02-01
Adding troubleshooting steps for unknown glyps icon in the prompt.

MrMeemus:
Due to the inconsistency of reproducing #111, I am providing a trouble shooting section on the usage guide.


doc/troubleshooting_guide.md
```diff
@@ -0,0 +1,72 @@
...
+
+## <a name="Usage_Issues"></a>Usage Issues:
+
+### Unknown glyph fills up prompt:
+If you encounter the display below, it is a Windows 10 issue that can pop up on the command prompt or powershell prompt.
+The current workaround for this issue is to change the font of the prompt to Consolas.
+![alt text](https://github.com/dbcli/mssql-cli/tree/master/screenshots/mssql-cli-display-issue.png "mssql-cli display issue")
+
+
+### <a name="Reporting_Issues"></a>Reporting Issues:
```
pensivebrian: Can we remove the anchor tags...they aren't needed

doc/troubleshooting_guide.md
```diff
@@ -0,0 +1,72 @@
...
+If the issue you are encountering is not listed above nor filed on our github, please file a issue here [issues](#https://github.com/dbcli/mssql-cli/issues) with the following information listed below and any other additional symptoms of your issue.
```
pensivebrian: Let's add 'which pip' and 'which python' to the commands to report.

pensivebrian: Can we also have the user post the commands used to install mssql-cli?

MrMeemus: Done.

doc/troubleshooting_guide.md
MrMeemus: I think we need the anchors to reference our headers since they have spaces in between them.

PR #155, opened 2018-02-02
Add special commands and tests

abhisheksinha89:
Adds the first batch of special commands, and the corresponding tests. 
Code coverage bumped from 63% to 64%

More special commands will be added in separate PR's.
This PR gives a foundation for implementing future special commands.

### DIFF
--- a/tests/test_special.py
+++ b/tests/test_special.py
@@ -1,21 +1,30 @@
 import unittest
-
+import uuid
 from mssqlutils import create_mssql_cli_client, shutdown
 from mssqlcli.packages.special.main import special_command, execute, NO_QUERY
 
+
 # All tests require a live connection to a SQL Server database
 class SpecialCommandsTests(unittest.TestCase):
+    session_guid = str(uuid.uuid4().hex)
+    table1 = 'mssql_cli_table1_{0}'.format(session_guid)
+    table2 = 'mssql_cli_table2_{0}'.format(session_guid)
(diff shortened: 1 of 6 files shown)


tests/test_special.py
```diff
@@ -0,0 +1,86 @@
...
+from mssqlcli.packages.special.main import special_command, execute, NO_QUERY
+
+# All tests require a live connection to a SQL Server database
+class SpecialCommandsTests(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        try:
+            # create the database objects to test upon
+            client = create_mssql_cli_client()
+            list(client.execute_single_batch_query('CREATE DATABASE testdb1;'))
```
pensivebrian: >testdb1 [](start = 68, length = 7)

can this by dynamic, eg. mssql-cli-test-xxxx-xxxxx

This way, concurrently running tests can't step on each other.

tests/test_special.py
```diff
@@ -0,0 +1,86 @@
+import unittest
+
```
pensivebrian: please run these test on supported server version to make sure the queries work: 2005, 2008, 2012, 2016, 2017, and Azure SQLDB, Azure SQL DW

abhisheksinha89: tested

---
In reply to: [165537064](https://github.com/dbcli/mssql-cli/pull/155#discussion_r165537064) [](ancestors = 165537064)

Issue #103, opened 2017-12-13
Results from completion refresher can leak onto the prompt


MrMeemus:
Repro steps:
Submit a single query multiple times in succession quickly and eventually a table from the completion refresher query will show up on the prompt.

Potential fix:
This could be related to client pulling off the completion refreshers response from the queue because they could share the same request ID. Will need more investigation.

MrMeemus:
Fixed in #161 .

Issue #95, opened 2017-12-13
A query is already in progress for this editor session


zayscue:
I'm running the latest release on windows 10 and when every I try to run a query on either single or multi line mode the only message I get is as follows, "A query is already in progress for this editor session. Please cancel this query or wait for its completion."  The problem is that I don't no other queries running.

MrMeemus:
@zayscue what kind of Server and DB are you targeting?

zayscue:
Microsoft SQL Server 2014 - 12.0.4100.1 (X64)   Apr 20 2015 17:29:27   Copyright (c) Microsoft Corporation  Developer Edition (64-bit) on Windows NT 6.3 <X64> (Build 9600: ) (Hypervisor)

MrMeemus:
I was able to repro this issue, it seems to be some concurrency issue from the back end. Will investigate more. Thanks @zayscue !

MrMeemus:
Dug a little deeper into this issue. It seems to be because the same mssqlcliclient is being used for completion refresher which may return results slowly. This can lead to the message you are seeing. This was originally by design but after fiddling with slower connections to DB's like Azure it seems like we should probably create a separate connection for the completion refresher. Will update this thread after further discussion.

kabirbaidhya:
I keep getting this issue too frequently almost every time I run query on a remote server. Do we have a plan of when this bug is going to get fixed and released?

MrMeemus:
@kabirbaidhya, I will be looking into this fix either late this week or early next week.

MrMeemus:
This issue has been fixed in #161. It should be available in our next release (0.10.0) ETA week of Feb 12th.

PR #161, opened 2018-02-08
Fix completion refresher query results leak and active query error

MrMeemus:
Fix for #103 and #95.

**a.)  A query is already in progress for this editor session. Please cancel this query or wait for its completion.**

Sqltoolsservice was not designed for MARS (Multiple Active Result Set) support. Meaning it cannot handle concurrent queries for a single connection. The previous design for mssql-cli was to share a connection between the main thread and the completion refresher thread. Due to this, if a query from the main thread was executed at the same time as the query from the background, this error message would be populated. 

**Solution:** Use a separate connection for the completion refresher thread. For Cloud support, this connection is new every time we refresh the completions when the database is altered or context changed.

**b.) Query results from background completion refresher thread leaks onto main prompt.**

When retrieving events from the shared JSON RPC client, a query/complete event would have no correlation with the request that submitted it. If the main thread executes in between the background threads query and fills the event queue with a query complete event, the background thread will break and never retrieve it's own results. The main thread then resumes and picks up the response from the background thread.

**Solution:** On top of maintaining a separate connection, responses are now retrieved via both a ID and/or Owner URI. This prevents connection A from retrieving results from connection B.

Added unit test:
- Verifies query A does not retrieve responses from query B.

Additional fixes:
- Logging for the completion refresher database connection.
- Not failing the completion refresher when it cannot connect.
- Correct casing for the error message event for the ResultSubset event.
- Additional logging to expedite future issues.
- Refactored and converged all JSON RPC contracts test into a single file as most of the code was redundant and hard to maintain.
- Add'tl test changes mentioned in the commits.


mssqlcli/main.py
```diff
@@ -594,14 +595,40 @@ def refresh_completions(self, history=None, persist_priorities='all'):
...
+                                          history=history,
+                                          settings=self.settings)
+
+
+
         return [(None, None, None,
                  'Auto-completion refresh started in the background.')]
 
+    def get_completion_refresher_mssqlcliclient(self, mssqlcliclient):
+        mssqlcliclient_completion_refreshser = MssqlCliClient(
```
pensivebrian: nit: naming - since we're creating a new MssqlCliClient, can we call this method create_completion_refresher_mssqlcliclient

pensivebrian: >mssqlcliclient_completion_refreshser [](start = 8, length = 36)

spelling: refreshser

pensivebrian: >MssqlCliClient [](start = 47, length = 14)

Perhaps we introduce a copy method to MssqlCliClient to centralize copying an instance?

mssqlcli/jsonrpc/contracts/tests/test_json_rpc_contracts.py
```diff
@@ -0,0 +1,249 @@
...
+            request = connectionservice.ConnectionRequest(id=u'1',
+                                                          owner_uri=owner_uri,
+                                                          json_rpc_client=rpc_client,
+                                                          parameters=parameters)
+            self.verify_connection_service_response(request=request,
+                                                    expected_response_event=1,
+                                                    expected_complete_event=1)
+            rpc_client.shutdown()
+
+    def test_query_execute_response_AdventureWorks2014(self):
```
pensivebrian: >test_query_execute_response_AdventureWorks2014 [](start = 8, length = 46)

Can we add a test that does concurrent queries on two different owner uri?

Also, can we add any tests for non-happy path failure scenarios? For example, the refresh connection is killed while executing while the query connection continues.

pensivebrian:
This PR is awesome! I can't wait to take this for a spin. Just some minor issues.

mssqlcli/main.py
MrMeemus: I originally thought that would be a good idea, but then OOP wise I don't think a mssqlclient should be responsible for copying itself and generating a new owner URI. I moved that responsibility up a level. Let me know your thoughts.

mssqlcli/jsonrpc/contracts/tests/test_json_rpc_contracts.py
MrMeemus: That's what this test does. It ensures that Request A never retrieves responses for Request B. This was the root cause of the display leak which is why I targeted this type of unit test. I can add a query connection failed test.

mssqlcli/main.py
MrMeemus: Fixed.

MrMeemus: Fixed.

pensivebrian: Two options: 

1) we currently generate an owner_id in the constructor, so can generate a new owner_id as part of the copy operation. 

2) we can pass in the owner_id in the copy method.

I think I like #2, and if the value not passed, we create it:

if not owner_uri:
            self.owner_uri = generate_owner_uri()

Maybe clone would be a better name to avoid shallow/deep copy confusion.

---
In reply to: [167103472](https://github.com/dbcli/mssql-cli/pull/161#discussion_r167103472) [](ancestors = 167103472)

mssqlcli/jsonrpc/contracts/tests/test_json_rpc_contracts.py
MrMeemus: Based on our discussion for restructuring the test story, I'll hold off on the failed connection test for now.

mssqlcli/main.py
```diff
@@ -594,14 +595,28 @@ def refresh_completions(self, history=None, persist_priorities='all'):
...
+                                          callbacks=callback,
+                                          history=history,
+                                          settings=self.settings)
+
+
+
         return [(None, None, None,
                  'Auto-completion refresh started in the background.')]
 
+    def create_completion_refresher_mssqlcliclient(self, mssqlcliclient):
```
pensivebrian: >create_completion_refresher_mssqlcliclient [](start = 8, length = 42)

I think we can delete this method can just call clone directly.

MrMeemus:
@pensivebrian ship it ribbit :)

PR #166, opened 2018-02-14
Updating query for when we are targeting Azure.

MrMeemus:
Listing logins special command uses a query against a table that does not exist in Azure.
Added conditional logic to select from a database principals instead. This table should retrieve the same information if connected to master. Otherwise it will retrieve logins on the database level.


MrMeemus:
woot!

PR #169, opened 2018-02-15
Removing redundant method in connection service.

MrMeemus:
Connection service should maintain a single responsibility of supplying connection requests and handling responses. It should not output any error messages to the console as that should be the responsibility of the client (mssqlcliclient).

After reviewing the handle_connection_response method it only served as a logging/console writing function. This was discovered during bug bash that if any errors happened on the completion refresher thread, the error would be populated to the user which is not useful. 

The background thread for completion refresher is allowed to fail and should not interrupt the user's session if it does.

### DIFF
--- a/mssqlcli/jsonrpc/contracts/connectionservice.py
+++ b/mssqlcli/jsonrpc/contracts/connectionservice.py
@@ -1,6 +1,5 @@
 from mssqlcli.jsonrpc.contracts import Request
 
-import sys
 import logging
 
 logger = logging.getLogger(u'mssqlcli.connectionservice')
@@ -42,7 +41,7 @@ class ConnectionRequest(Request):
             return decoded_response
 
         except Exception as error:
-            logger.debug(str(error))
+            logger.info(str(error))
             self.finished = True
             self.json_rpc_client.request_finished(self.id)
             self.json_rpc_client.request_finished(self.owner_uri)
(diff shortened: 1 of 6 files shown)


mssqlcli/jsonrpc/contracts/connectionservice.py
```diff
@@ -144,38 +143,3 @@ class ConnectionResponse(object):
...
-                logger.error(response.messages)
-
-        return response
-
-    def handle_connection_response_notification(response):
-        if not response.result:
-            sys.stderr.write(
-                u'\nIncorrect json rpc request. Connection not successful.')
-        return None
-
```
abhisheksinha89: These output to the terminal connection error messages. Without these the user would'nt know why the connection did not succeed.

MrMeemus: Sure, I can surface error messages for the client to output to the console. For this particular error how do we know the lack of a result means the json rpc request was invalid? also this message doesn't seem to be actionable for the user since we are forming the request.

mssqlcli/mssqlcliclient.py
```diff
@@ -99,11 +99,15 @@ def connect(self):
...
             response = connection_request.get_response()
-            if response:
-                response = connectionservice.handle_connection_response(
-                    response)
+
+            if isinstance(response, connectionservice.ConnectionCompleteEvent):
+                if response.error_message:
+                    error_messages.append(u'Error message: {}'.format(response.error_message))
+                if response.messages:
+                    error_messages.append(u'Additional message: {}'.format(response.messages))
```
abhisheksinha89: Do not add additional messages. This has the entire call stack of SqlClient. That is why previously we just output the error and only logged the message.

abhisheksinha89:
Did you try this with an incorrect login to see what is output to the user?

mssqlcli/mssqlcliclient.py
MrMeemus: updated.

mssqlcli/jsonrpc/contracts/connectionservice.py
abhisheksinha89: We would get back a result if toolsservice was able to understand our request. If we don't have a result parameter it means something went wrong in jsonrpc request. 

---
In reply to: [168625939](https://github.com/dbcli/mssql-cli/pull/169#discussion_r168625939) [](ancestors = 168625939)

mssqlcli/jsonrpc/contracts/tests/test_json_rpc_contracts.py
```diff
@@ -173,8 +173,6 @@ def verify_connection_service_response(self,
                 elif isinstance(response, connectionservice.ConnectionCompleteEvent):
                     if response.connection_id:
                         complete_event += 1
```
abhisheksinha89: can you assert the connection_id here?

abhisheksinha89: nvm

---
In reply to: [168636959](https://github.com/dbcli/mssql-cli/pull/169#discussion_r168636959) [](ancestors = 168636959)

PR #170, opened 2018-02-23
Partial refactor of code base

MrMeemus:
Changes made:

**Argument handling**
- Moved from click to argparse. Still retaining click for it's echo via pager API.
- All objects are accessed via the namespace object generated by argparser.
- Arg parse can handle defaults better, no longer requiring explicit checks to set another option. 
   Example: If no server is passed in, we automatically target master. This reduced checks and centralizes all options into a single file.

**Tests**
- Tests have been updated to handle options via a namespace object. This allows each test method to express what options to toggle along with retaining default values. 
 - All mssql client and option creations are centralized and no longer need explicitly set values on all calls unless a non default option is necessary. This is evident in test_mssqlcliclient.py and test_rowlimit.py

**Miscellaneous**
- Methods were renamed for readability.
- Added additional layers of abstraction for readability.
- Messages logged are trimmed since some did not provide additional value.
- Updated mssqlcliclient to support a shallow clone method.
- Mssqlcliclient has private methods added to abstract away the connection and query execution API.
- sqltoolsclient API are expressed in a class variable so as to avoid any hardcoded JSON RPC calls.
- lower case all action for build.py for robustness.

### DIFF
--- a/build.py
+++ b/build.py
@@ -141,8 +141,9 @@ if __name__ == '__main__':
     actions = sys.argv[1:] or default_actions
 
     for action in actions:
-        if action in targets:
-            targets[action]()
+        current_action = action.lower()
+        if current_action in targets:
+            targets[current_action]()
         else:
-            print('{0} is not a supported action'.format(action[1]))
+            print('{0} is not a supported action'.format(current_action[1]))
             print('Supported actions are {}'.format(list(targets.keys())))
--- a/build.py
+++ b/build.py
@@ -128,6 +128,14 @@ def test():
     integration_test()
 
 
+def validate_actions(user_actions, valid_targets):
+    for user_action in user_actions:
+        if user_action.lower() not in valid_targets.keys():
+            print('{0} is not a supported action'.format(user_action))
+            print('Supported actions are {}'.format(list(valid_targets.keys())))
+            sys.exit(1)
+
+
 if __name__ == '__main__':
     default_actions = ['build', 'unit_test']
 
@@ -140,10 +148,7 @@ if __name__ == '__main__':
     }
     actions = sys.argv[1:] or default_actions
 
+    validate_actions(actions, targets)
+
     for action in actions:
-        current_action = action.lower()
-        if current_action in targets:
-            targets[current_action]()
-        else:
-            print('{0} is not a supported action'.format(current_action[1]))
-            print('Supported actions are {}'.format(list(targets.keys())))
+        targets[action]()
--- a/mssqlcli/main.py
+++ b/mssqlcli/main.py
@@ -58,9 +58,9 @@ except ImportError:
 from collections import namedtuple
 
 # mssql-cli imports
-from mssqlcli.mssqlclioptionsparser import get_parser
+from mssqlcli.mssqlclioptionsparser import create_parser
 from mssqlcli.sqltoolsclient import SqlToolsClient
-from mssqlcli.mssqlcliclient import MssqlCliClient, reset_connection_and_clients
+from mssqlcli.mssqlcliclient import MssqlCliClient, generate_owner_uri
 from mssqlcli.packages import special
 import mssqlcli.telemetry as telemetry_session
 
@@ -316,8 +316,7 @@ class MssqlCli(object):
             # Right now the sql_tools_service process is killed and we restart
             # it with a new connection.
             click.secho(u'Cancelling query...', err=True, fg='red')
-            reset_connection_and_clients(self.sqltoolsclient,
-                                         self.mssqlcliclient_main)
+            self.reset_connection()
             logger.debug("cancelled query, sql: %r", text)
             click.secho("Query cancelled.", err=True, fg='red')
 
@@ -505,7 +504,7 @@ class MssqlCli(object):
             exit(1)
 
         for rows, columns, status, sql, is_error in \
-                self.mssqlcliclient_main.execute_multi_statement(text):
+                self.mssqlcliclient_main.execute_query(text):
             total = time() - start
             if self._should_show_limit_prompt(status, rows):
                 click.secho('The result set has more than %s rows.'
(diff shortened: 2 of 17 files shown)


build.py
```diff
@@ -141,8 +141,9 @@ def test():
     actions = sys.argv[1:] or default_actions
 
     for action in actions:
-        if action in targets:
-            targets[action]()
+        current_action = action.lower()
+        if current_action in targets:
+            targets[current_action]()
```
pensivebrian: I think we should validate all actions and fail before we execute any action.

mssqlcli/mssqlclioptionsparser.py
```diff
@@ -0,0 +1,159 @@
...
+
+MSSQL_CLI_USER = u'MSSQL_CLI_USER'
+MSSQL_CLI_PASSWORD = u'MSSQL_CLI_PASSWORD'
+MSSQL_CLI_DATABASE = u'MSSQL_CLI_DATABASE'
+MSSQL_CLI_SERVER = u'MSSQL_CLI_SERVER'
+MSSQL_CLI_ROW_LIMIT = u'MSSQL_CLI_ROW_LIMIT'
+MSSQL_CLI_RC = u'MSSQL_CLI_RC'
+
+
+def initialize():
```
pensivebrian: We have two public methods that do the same thing: initialize and get_parser. Can we have a single method: create_parser?

mssqlcli/mssqlclioptionsparser.py
```diff
@@ -0,0 +1,159 @@
...
+    args_parser.add_argument(
+        u'--prompt',
+        dest=u'prompt',
+        metavar=u'',
+        help=u'Prompt format (Default: \\d> ')
+
+    return args_parser
+
+
+def get_parser():
```
pensivebrian: get_parser implies we are returning an existing instance. Since were creating a new one each call, I'd prefer we name this create_parser or build_parser.

mssqlcli/mssqlcliclient.py
```diff
@@ -335,19 +342,19 @@ def reset_connection_and_clients(sql_tools_client, *mssqlcliclients):
...
+            mssqlcli_client.is_connected = False
+            mssqlcli_client.owner_uri = generate_owner_uri()
+            if not mssqlcli_client.connect_to_database():
+                click.secho('Unable reconnect to server {0}; database {1}.'.format(mssqlcli_client.server_name,
+                                                                                   mssqlcli_client.database),
+                            err=True, fg='yellow')
+                logger.info(u'Unable to reset connection to server {0}; database {1}'.format(
+                    mssqlcli_client.server_name,
+                    mssqlcli_client.database))
                 exit(1)
```
pensivebrian: I think this logic can be moved into a reset instance method.

mssqlcli/mssqlcliclient.py
```diff
@@ -335,19 +342,19 @@ def reset_connection_and_clients(sql_tools_client, *mssqlcliclients):
     try:
         sql_tools_client.shutdown()
         new_tools_client = SqlToolsClient()
```
pensivebrian: A client should not reinitialize the sqltoolsservice process. Can we move this a reset method in the sqltoolsclient client?

mssqlcli/mssqlcliclient.py
```diff
@@ -23,83 +23,143 @@ def generate_owner_uri():
...
+    def connect_to_database(self):
+        connection_params = self.get_base_connection_params()
+        connection_params = self.add_optional_connection_params(connection_params)
+
+        owner_uri, error_messages = self._execute_connection_request_with(connection_params)
+
+        return owner_uri, error_messages
+
+    def execute_multi_statement(self, query):
+        # Try to run first as special command
```
pensivebrian: From and API standpoint, it would be nice if we had a single public method to execute on or more statements: execute_query. So execute_single_statement would become private. Any downside to this?

pensivebrian:
> class MssqlCli(object):

What do you think about moving MssqlCli into it's own file?

---
Refers to: mssqlcli/main.py:102 in 4eb4555. [](commit_id = 4eb4555268f2cca9066261808de811bd069f2ccb, deletion_comment = False)

pensivebrian:
>     """

I think this operation should be moved to the MssqlCli object.

---
Refers to: mssqlcli/mssqlcliclient.py:338 in 4eb4555. [](commit_id = 4eb4555268f2cca9066261808de811bd069f2ccb, deletion_comment = False)

MrMeemus:
@pensivebrian , YES I WANT TO MOVE MSSQLCLI INTO IT's OWN FILE :) , I can do this on the side as I'm prioritizing the packaging work now.

mssqlcli/mssqlcliclient.py
MrMeemus: Agreed, no downsides other then background queries will run through additional code path but it wouldn't be harmful.

---
In reply to: [170340117](https://github.com/dbcli/mssql-cli/pull/170#discussion_r170340117) [](ancestors = 170340117)

mssqlcli/main.py
```diff
@@ -623,13 +559,38 @@ def _handle_server_closed_connection(self):
...
-                reset_connection_and_clients(self.sqltoolsclient,
-                                             self.mssqlcliclient_query_execution)
+                self.reset_connection()
 
                 click.secho('Reconnected!\nTry the command again.', fg='green')
             except Exception as e:
                 click.secho(str(e), err=True, fg='red')
 
+    def reset_connection(self):
+        """
```
pensivebrian: Maybe just call this reset?

pensivebrian: If you change this, please update the method name in the logging below.

---
In reply to: [170392798](https://github.com/dbcli/mssql-cli/pull/170#discussion_r170392798) [](ancestors = 170392798)

mssqlcli/main.py
```diff
@@ -623,13 +559,38 @@ def _handle_server_closed_connection(self):
...
+        Reset mssqlcli client with a new sql tools service and connection.
+        """
+        try:
+            self.sqltoolsclient.shutdown()
+            self.sqltoolsclient = SqlToolsClient()
+
+            self.mssqlcliclient_main.sql_tools_client = self.sqltoolsclient
+            self.mssqlcliclient_main.is_connected = False
+            self.mssqlcliclient_main.owner_uri = generate_owner_uri()
+
```
pensivebrian: I feel like we should clone the existing mssqlclient such that we know we're in a clean state.

MrMeemus: agreed, I like that idea. Will add it next.

---
In reply to: [170394199](https://github.com/dbcli/mssql-cli/pull/170#discussion_r170394199) [](ancestors = 170394199)

pensivebrian:
:shipit:  Awesome! I really like how things cleaned things up.

abhisheksinha89:
Can we file a GitHub issue to completely remove click once we are done with refactoring?

mssqlcli/main.py
MrMeemus: It sounds more intuitive to have reset_connection because that is explicitly resetting, what do you think? 

Having just reset can imply a lot of options to be reset for MssqlCli.

---
In reply to: [170392798](https://github.com/dbcli/mssql-cli/pull/170#discussion_r170392798) [](ancestors = 170392798)

MrMeemus:
@abhisheksinha89  https://github.com/dbcli/mssql-cli/issues/171

PR #173, opened 2018-03-01
Ron/mssqlcli refactor

MrMeemus:
- Moved MssqlCli into it's own file (__init__.py).
- Created a mssqlutils module that has the static methods used by MssqlCli.
- Removed unused static methods such as is_mutating_cmd, etc.
- Renamed mssqlutils that was used by tests to mssqltestutils.
- Moved telemetry session server log to main.py as MssqlCli should not be aware of a telemetry object.


mssqlcli/__init__.py
```diff
@@ -1 +1,638 @@
...
+    'OutputSettings',
+    'table_format dcmlfmt floatfmt missingval expanded max_width case_function'
+)
+OutputSettings.__new__.__defaults__ = (
+    None, None, None, '<null>', False, None, lambda x: x
+)
+
+
+class MssqlCli(object):
+
```
pensivebrian: Do we have to put this in __init__.py? I'd rather see it in mssqlcli.py

abhisheksinha89: Yea putting stuff in init.py is misleading IMO. Agree with Brian

---
In reply to: [171445777](https://github.com/dbcli/mssql-cli/pull/173#discussion_r171445777) [](ancestors = 171445777)

mssqlcli/mssqlutils.py
```diff
@@ -0,0 +1,110 @@
...
+        if first_token.lower() in ('alter', 'create', 'drop'):
+            return True
+    except Exception:
+        return False
+
+    return False
+
+
+def has_change_db_cmd(query):
+    """Determines if the statement is a database switch such as 'use' or '\\c'
```
pensivebrian: My preference... if a method is only used in one file, define the method that file. In this case, has_changed_db_cmd is only used by MssqlCli, so I think it should be there.

pensivebrian: This may apply to other methods in this file.

---
In reply to: [171446837](https://github.com/dbcli/mssql-cli/pull/173#discussion_r171446837) [](ancestors = 171446837)

MrMeemus:
Refactored based on comments. I moved MssqlCli into mssql_cli.py because when I named it mssqlcli there was a issue with telelemtry upload where it would try to resolve mssqlcli.config from the module rather than the package level. We can refactor this later, it's just not as a high priority for now.

PR #174, opened 2018-03-02
Ron/file history change

MrMeemus:
Any statement typed that contains the keywords 'secret' or 'password' is not stored and saved in the query file history.

The feature that autosuggests on the prompt is based on the contents of the file history. By not adding the flagged statements, we avoid a suggestion of that statement.


mssqlcli/mssql_cli.py
```diff
@@ -667,6 +685,16 @@ def has_change_db_cmd(self, query):
 
         return False, None
 
+    def has_security_statement_in_cmd(self, query):
+        """Determines if the statement contains a PASSWORD or SECRET keyword
+        """
+        try:
+            tokens = query.lower()
+            return 'password' in tokens or 'secret' in tokens
```
amjith: Are these the only tokens we want to ignore? Is there any value in making a list of forbidden words and iterating over them to check if any tokens match that?

MrMeemus: ah yes, we can have a top level list that both MssqlCliFileHistory and this function can reference.

mssqlcli/mssql_cli.py
```diff
@@ -667,6 +687,17 @@ def has_change_db_cmd(self, query):
 
         return False, None
 
+    def has_security_statement_in_cmd(self, query):
+        """Determines if the statement contains a PASSWORD or SECRET keyword
+        """
+        try:
+            tokens = query.lower()
+            keywords_flagged = [keyword for keyword in security_keywords if keyword in tokens]
+            return keywords_flagged
```
pensivebrian: Isn't this the same logic defined in the MssqlFileHistory class? It'd be nice to put the logic in one place, as simple as the logic may be.

mssqlcli/mssql_cli.py
```diff
@@ -73,6 +74,22 @@
     None, None, None, '<null>', False, None, lambda x: x
 )
 
+security_keywords = ['password', 'secret']
+
```
pensivebrian: We should add ENCRYPTED_VALUE

mssqlcli/mssql_cli.py
```diff
@@ -489,6 +507,8 @@ def _evaluate_command(self, text):
                     click.secho("Aborted!", err=True, fg='red')
                     break
 
+            contains_secure_statement = self.has_security_statement_in_cmd(sql)
+
```
pensivebrian: perf nit: aren't we running the function twice? Once for file history and once here executing the command? It'd be nice todo it once, though it might not make sense if it complicates the code.

pensivebrian:
Looks good! Just a couple minor comments.

mssqlcli/mssql_cli.py
MrMeemus: The file history is tied to prompt tool kitwhich attempts to load the previous type statement if the first letter matches it. This code block is for preventing the smart completion from loading the statements.

mssqlcli/mssql_cli.py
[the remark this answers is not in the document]
MrMeemus: that looks much cleaner, thanks for the suggestion.

MrMeemus: I'll define it as a module level method that both functions can reference.

mssqlcli/mssql_cli.py
pensivebrian: Understood. But the question still stands, aren't we parsing the same query twice?

tests/test_main.py
```diff
@@ -10,7 +11,25 @@
...
+    ]
+
+    with tempfile.TemporaryFile(delete=False) as temp_file:
+        temp_file_path = temp_file.name
+        file_history = MssqlFileHistory(temp_file_path)
+
+        for statement in statements:
+            file_history.append(statement)
+
+        assert len(file_history) == 0
```
pensivebrian: Should we be checking the length of the temp file here?

mssqlcli/mssql_cli.py
MrMeemus: Yes, I can add a flag in the filehistory object that we can check if the result query contained security keywords. Let me know your thoughts.

PR #181, opened 2018-04-04
Debian and Rpm packaging scripts

MrMeemus:
This PR adds builds scripts used to generate .rpm and .deb packages for mssql-cli.

- Both packages will contain their copy of python 3.6 bundled in.
- The versioning schema for daily builds will follow the same as the daily whls,  0.10.0.dev<timestamp>
- Package building has been verified on Ubuntu 14 - 17 and redhat 
- Dependency resolution has been verified to be handled by apt-get or yum. In particular our debian package contains alternative dependencies which resolve the issue between the different version of libicu per major version of Ubuntu.


setup.py
```diff
@@ -1,14 +1,10 @@
...
-# Licensed under the MIT License. See License.txt in the project root for license information.
-# --------------------------------------------------------------------------------------------
-
 import ast
 import os
 import re
 import datetime
 
 from setuptools import setup, find_packages
+from codecs import open
```
pensivebrian: Why is this not needed?

pensivebrian:
> Microsoft Privacy statement: https://privacy.microsoft.com/en-us/privacystatement

Not part of your CR, but can you remove the hard coded en-us path from the url? It we auto-redirect to the correct locale based on the users browser language settings.

---
Refers to: mssqlcli/main.py:36 in 0d624cb. [](commit_id = 0d624cb1954f4960a3b4209be0684797161d7298, deletion_comment = False)

setup.py
MrMeemus: During the build process for RPM, rpm detects that it is a python package and uses it's built in python 2.7 to clean the build. This fails when it hits the long description of setup.py with a unicode decode error. The call to open requires a encoding utf-8 flag but is not supported on python 2. So instead we import built in codecs which handles the encoding for us.

pensivebrian: Interesting! Thanks.

build_scripts/rpm/mssql-cli.spec
```diff
@@ -0,0 +1,86 @@
...
+%description
+    We’re excited to introduce mssql-cli, a new and interactive command line query tool for SQL Server.
+    This open source tool works cross-platform and proud to be a part of the dbcli.org community.
+
+%prep
+# Clean previous build directory.
+rm -rf %{_builddir}/*
+# Download, Extract Python3
+python_archive=$(mktemp)
+wget https://www.python.org/ftp/python/3.6.1/Python-3.6.1.tgz -qO $python_archive
```
meeuw: Don't use wget in `%prep`, you should define the url in `source` and you can use `spectool -g` to fetch it.

MrMeemus: What are the drawbacks of using wget?

MrMeemus:
Windows build failed because the build def had "fail on standard error" which when pip is out of date seems to be a error. Reran build and verified it was successful.

PR #183, opened 2018-04-04
Update daily builds to drop in daily/whl

MrMeemus:
Debian and RPM packages will drop in daily/deb and daily/rpm respectively.


release.py
```diff
@@ -6,7 +6,7 @@
 
 AZURE_STORAGE_CONNECTION_STRING = os.environ.get(
     'AZURE_STORAGE_CONNECTION_STRING')
-BLOB_MSSQL_CLI_DAILY_CONTAINER_NAME = 'daily/mssql-cli'
+BLOB_MSSQL_CLI_DAILY_CONTAINER_NAME = 'daily/whl'
```
pensivebrian: I though we needed the package name in the path for pip. Please confirm it works with the --extra-index-url parameter.

MrMeemus:
@pensivebrian yes, there are also changes for the deb rpm packaging which I am waiting for review on so I can update the readme/installation guide all at once.

MrMeemus:
Pip install with extra index has been verified. Installation guide will be updated in a separate PR for all packaging formats we have.

PR #188, opened 2018-04-17
Installation docs update

MrMeemus:
Docs have been updated to support linux packaging install story.
Most links will be broken until we merge.


README.rst
```diff
@@ -67,7 +107,6 @@ mssql-cli is licensed under the `BSD-3 license`_.
 
 .. _mssql-cli: https://github.com/dbcli/mssql-cli
 .. _dbcli.org community: https://github.com/dbcli
-.. _installation guide: https://github.com/dbcli/mssql-cli/tree/master/doc/installation_guide.md
```
pensivebrian: This is probably linked to from external sources, so I don't think we should delete it. I think we should either move that table from the README here. Or figure out how to put something useful here.

MrMeemus: Sure, for now I'll add a blurb to move to new link for now.

PR #191, opened 2018-04-30
Telemetry updates

MrMeemus:
- Daily user id rotation: We generate a new cryptographically secure id for the user for each 24 hour window of usage.

- Moved import statement to prevent creation of config dir prior to check so we can display telemetry message.

- Added envelope flag to suppress ingest header for vortex.

### DIFF
--- a/mssqlcli/telemetry.py
+++ b/mssqlcli/telemetry.py
@@ -1,3 +1,4 @@
+import binascii
 import datetime
 import json
 import locale
@@ -8,7 +9,9 @@ import sys
 import traceback
 import uuid
 from functools import wraps
-
+from datetime import datetime as temp
+from datetime import timedelta
+import datetime
 import mssqlcli.config as config
 import mssqlcli.telemetry_upload as telemetry_core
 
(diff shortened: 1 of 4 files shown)


mssqlcli/telemetry.py
```diff
@@ -170,23 +173,33 @@ def _get_mssql_cli_version():
...
+        return id_file_creation_time < last_24_hours
+    return False
 
 
 @decorators.suppress_all_exceptions(fallback_return='')
 @decorators.hash256_result
 def _generate_user_id():
-    random_id = uuid.uuid4().hex
-    salt = uuid.uuid4().hex
+    random_id = binascii.hexlify(os.urandom(16)).decode() \
```
pensivebrian: We can combine this into one call to os.urandom(32) since the id/salt distinction doesn't really apply anymore?

mssqlcli/telemetry.py
```diff
@@ -1,3 +1,4 @@
+import binascii
 import datetime
```
pensivebrian: Would be nice to add a test that validate the file is updated each 24 hrs.

mssqlcli/telemetry.py
MrMeemus: good point, will do.

PR #196, opened 2018-05-08
Update twine from 1.8.1 to 1.11.0

MrMeemus:
During publish step we encounter this issue:
https://github.com/pypa/twine/issues/342

Updating twine is the fix.

Added missing colon to default prompt for user name.

Added format fix for version display on prompt.


mssqlcli/main.py
```diff
@@ -63,7 +63,7 @@ def configure_and_update_options(options):
 
     if not options.integrated_auth:
         if not options.username:
-            options.username = input(u'Username (press enter for sa)') or u'sa'
+            options.username = input(u'Username (press enter for sa):') or u'sa'
```
pensivebrian: PR of the month right here! This has bugged me for so long :)

abhisheksinha89: My mistake is finally corrected

PR #199, opened 2018-05-10
Update usage_guide.md

bigleka:
Change a information about the parameter option from -E to -I for integrated authentication on Windows


MrMeemus:
@bigleka good catch! thanks :)

PR #200, opened 2018-05-10
Fixing attribute error for column meta data.

MrMeemus:
A previous refactor replaced the call with get_foreign_keys. This was missed by tests as we don't have a test suite up for the completion engine just yet. 

The test gap will be addressed in the near future.


jefflomax:
Verified.

Issue #189, opened 2018-04-29
Authentication to AD/LDAP-accounts


anton-johansson:
I have a server with authentication through active directory. But I can't seem to login. If I were on Windows, I suppose it would work with the `--integrated` switch. But how would I do on a Linux platform?

pensivebrian:
To use integrated authentication with mssql-cli on macOS or Linux, you first have to configure Kerberos using the [following steps](https://github.com/Microsoft/vscode-mssql/blob/master/KERBEROS_HELP.md). The steps are a bit tedious, but once configured, you should be able to connect using the `-E` or `--integrated` switch.

chucker:
Do you mean `-I`? That yields:

>Integrated authentication not supported on this platform

Even if I previously ran `kinit`.

ScriptMyJob:
Yeah, this is super annoying.  I am surprised that `mssql-cli` does not seem to support domain authentication on MacOS.  I seem to remember that I got this working in an earlier build (using `sqlcmd` on MacOS) but I am damned if I cannot figure it out now.  ¯\_(ツ)_/¯

There really should be a switch to handle AD Authentication that is not Azure.

geoffclark:
Hi I cannot get mssql-cli to use my AD\LDAP account
It is correctly set up as I can use AD Authentication for SQL Operations Studio 2
Any help would be appreciated.

djustins:
using RHEL7, also get:
Integrated authentication not supported on this platform

Even though i have a successful krb ticket

pensivebrian:
We'll need to investigate this.

@MrMeemus - we should be using the -E switch for integrated auth to align with sqlcmd.

MrMeemus:
There was a method that previously blocked integrated auth on non windows platforms due to the limitation in .Net Core prior. This has been fixed in #214 and will be slated for 0.15.0 release. Thanks all!

PR #214, opened 2018-07-02
Removing blocker when using integrated auth on Mac/Linux 

MrMeemus:
- Also renaming parameter for integrated auth from -I to -E to align with SqlCMD and what it was originally.


MrMeemus:
Resolves #189.

Issue #227, opened 2018-10-21
use of `scripts` in setup.py causes errors with virtualenvs


cs01:
When running `mssql-cli` from within an activated virtualenv, everything is fine because the virtual env's `python` is the first match on the PATH. However, when invoking `mssql-cli` from `myvenv/bin/mssql-cli` while the virtual environment is **not** activated, we get
```
> myvenv/bin/mssql-cli
/usr/bin/python: Error while finding module specification for 'mssqlcli.main' (ModuleNotFoundError: No module named 'mssqlcli')
```
as noted in https://github.com/cs01/pipx/issues/12.

This is because the script `mssql-cli` is custom made and hardcoded to invoke a new bash shell, ignoring any virtual environment. 

There are a few solutions. The first is easiest and most common.
1. Use the standard `entry_points` entry instead of `scripts`
```
    entry_points = {"console_scripts": [
        "mssql-cli.bat = mssqlcli.main:main",
        "mssql-cli = mssqlcli.main:main"
        ]},
```
2. Leave as a script, but use python instead of bash; make the interpreter `/usr/bin/env python` instead of `bash`, similar to how awscli does it: https://github.com/aws/aws-cli/blob/develop/bin/aws
3. In the script `mssql-cli` test if there is an `activate` script present, and if so, run `source activate`  before running `python -m mssqlcli.main "$@"`. This is hacky and I would not go this route.

MrMeemus:
@cs01  thanks for being so proactive with this issue! The only concern I had is if we are not referencing the scripts any more we lose the IO Encoding. This would affect us when we support pipeable output, which we can revisit when we cross that bridge. 

I'll go ahead and merge your change. Thanks again!

cs01:
I just realized the two scripts that are no longer used should be deleted as well. I didn't do that in the PR.

cs01:
> One question I do have, what is the expected behavior if running with multiple virtual environments that are not activated?

Sorry, I never responded to this. If a virtual environment is not activated, then these entry points are not available on your $PATH. 

>  or the case when another virtual environment is activated that did not install mssql-cli?

If this is the case, then the entry points are also not available on your $PATH.

PR #228, opened 2018-10-21
replace setup.py scripts with console entry_points #227

cs01:
https://github.com/dbcli/mssql-cli/issues/227


nickolay:
This was rolled back in https://github.com/dbcli/mssql-cli/pull/299

PR #229, opened 2018-10-26
remove unused scripts

cs01:


petobens:
Friendly ping :)

cs01:
Oh I forgot I had this diff up still. 

```
pipx install mssql-cli --spec git+https://github.com/dbcli/mssql-cli
```

works for me.

```
pipx install mssql-cli
```
might work for you, but I got an error.

petobens:
Oh ok. What I meant was that installing this with pipx works but I cannot use the binary thereafter: when I actually try to connect to a database with `mssql-cli -S host -U user -P password` I get the error above.

cs01:
Oh ok gotcha. Did you try
```
pipx install mssql-cli --spec git+https://github.com/dbcli/mssql-cli
```
too?

If it still fails, I'm not certain whether this is a bug in pipx or mssql-cli. I guess if it works without pipx (i.e. with a normal pip install), it's a pipx bug. Is this the case?

petobens:
Mmm `pip install --user mssql-cli` works fine but 
```
pip install --user git+https://github.com/dbcli/mssql-clii
```
gives the same error (the one I've posted above) as 
```
pipx install mssql-cli --spec git+https://github.com/dbcli/mssql-cli
```
Any ideas what might be going on?

nickolay:
This was closed without merging because similar changes were landed in https://github.com/dbcli/mssql-cli/pull/228

The fix has since been reverted, re-introducing pipx problems: https://github.com/dbcli/mssql-cli/issues/496

PR #234, opened 2018-12-13
moved to use python3, fixed install docs.

EvanCarroll:
There is no reason to have different install docs for each different Debian-based distro. Now we have one set of docs that should work with them all (include Ubuntu and Mint). Also made pip install dependant on pip3 for Debian based distros. Made the `mssql-cli` check for the existance of python3, before falling back to python. Python is not the "default" for Python2/3. Python3 should be the default.

### DIFF
--- a/doc/installation/linux.md
+++ b/doc/installation/linux.md
@@ -2,9 +2,7 @@
 
 Supports:
 
-- [Ubuntu 14.04][u14]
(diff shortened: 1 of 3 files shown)


MrMeemus:
@EvanCarroll , will take care of it today. Thanks!

PR #254, opened 2019-05-27
Upgrade MSSQL Tools Service

CrshOverride:
Currently, `mssql-cli` relies on 2.1 builds of the Sql Tools (which, in turn, require `libssql1.0`. This causes issues for users using recent releases of `libssl` (especially those on Ubuntu 19.04).

This PR updates the packages to more recent releases.


pensivebrian:
Thanks for the PR @CrshOverride, but unfortunately we didn't have the bandwidth to test and merge this. However we did updated the SqlToolsService with the 0.16.0 release.

PR #255, opened 2019-06-03
add ArchLinux steps

noraj:


pensivebrian:
Hi, @noraj. Though I appreciate the PR, I'm not going to merge this. Arch user are probably already familiar with how to search and install Arch packages, and I'm hesitant to point to a package repro that we don't maintain or support.

Issue #221, opened 2018-09-15
Wrong table completion after database change


thorstenkampe:
After switching databases with `use another_db` tab completion for tables - for instance after `select * from <TAB>` completes still from the tables of the previous database, not the current one.

Other members of the `dbcli` family complete tables from the current database.

yualan:
I repro-ed this as well, team will investigate

gdzwrk:
I am still experiencing this- if I connect using the `-d` option and specify my database, I receive proper table/view/FK completion suggestions.  If I omit the -d, by default it connects to the 'master' db.  If I then `use other_db` I am still only shown completion suggestions from master.

For now I am working around this by specifying the -d option when invoking the program.

PR #256, opened 2019-06-17
mssqlcliclient: When connecting, respect `connected_database`

detule:
Please consider this commit.

It fixes: https://github.com/dbcli/mssql-cli/issues/221
This problem was introduced in https://github.com/dbcli/mssql-cli/pull/215


pensivebrian:
Thanks, @detule. We really appreciate you taking the time to make this fix!

PR #258, opened 2019-07-27
Fixed broken unit tests: test_mssqlcliclient.py and test_completion_refresher.py

geleems:
Fixed broken unit tests (`test_mssqlcliclient.py` and `test_completion_refresher.py`)

Failed unit tests in `test_mssqlcliclient.py`
* test_schema_table_views_and_columns_query
  * Cause: `client.get_views()` failed to retrieve list of views since it uses query with where clause `WHERE table_type = 'View'` that returns nothing on case-sensitive database
  * Fix: Remove the where clause and look up `INFORMATION_SCHEMA.VIEWS` instead of `INFORMATION_SCHEMA.TABLES`

Failed unit tests in `test_completion_refresher.py`

* test_ctor
  * Causes:
    * `refresher.refreshers.keys()` returns only 5 refresher names defined in `completion_refresher.py`
    * `set` comparison should have been used instead of `list` comparison to verify if all elements of `A` are in `B` and all elements of `B` are in `A`
  * Fix
    * Update the `expected_handlers` with appropriate refreshers
    * Use `set` comparison instead of `list`

* test_refresh_with_callbacks
  * Cause: it used wrong mock object for `mssqlcliclient` which blows up when `mssqlcliclient.connect_to_database()` is called.
  * Fix: Create and use mock object for `mssqlcliclient` with `connect_to_database()` implementation.

Changed use unit test framework `unittest` since class type unit test that extends unittest.TestCase (such as test_mssqlcliclient.py) are a lot more convenient to run and manage.

### DIFF
--- a/mssqlcli/mssqlqueries.py
+++ b/mssqlcli/mssqlqueries.py
@@ -90,8 +90,7 @@ def get_views():
     return '''
         SELECT  table_schema,
                 table_name
-        FROM INFORMATION_SCHEMA.TABLES
-        WHERE table_type = 'View'
+        FROM INFORMATION_SCHEMA.VIEWS
         ORDER BY 1, 2'''
 
 
--- a/tests/test_completion_refresher.py
+++ b/tests/test_completion_refresher.py
@@ -1,87 +1,82 @@
 import time
-import pytest
+import unittest
 from mock import Mock, patch
+from mssqlcli.completion_refresher import CompletionRefresher
 
+class CompletionRefresherTests(unittest.TestCase):
 
-@pytest.fixture
-def refresher():
-    from mssqlcli.completion_refresher import CompletionRefresher
-    return CompletionRefresher()
+    def test_ctor(self):
+        """
+        Refresher object should contain a few handlers
+        :param refresher:
+        :return:
+        """
+        refresher = CompletionRefresher()
+        assert len(refresher.refreshers) > 0
+        actual_handlers = set(refresher.refreshers.keys())
+        expected_handlers = set(['databases', 'schemata', 'tables', 'types', 'views'])
+        assert expected_handlers == actual_handlers
 
+    def test_refresh_called_once(self):
+        """
 
-def test_ctor(refresher):
-    """
-    Refresher object should contain a few handlers
-    :param refresher:
-    :return:
-    """
-    assert len(refresher.refreshers) > 0
-    actual_handlers = list(refresher.refreshers.keys())
-    expected_handlers = ['schemata', 'tables', 'views',
-                         'types', 'databases', 'casing', 'functions']
-    assert expected_handlers == actual_handlers
+        :param refresher:
+        :return:
+        """
+        callbacks = Mock()
+        mssqlcliclient = Mock()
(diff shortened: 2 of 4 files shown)


tests/test_completion_refresher.py
```diff
@@ -1,87 +1,82 @@
 import time
-import pytest
+import unittest
```
pensivebrian: Is unittest a different unit test framework? If so, why not use pytest?

tests/test_completion_refresher.py
```diff
@@ -1,87 +1,82 @@
...
+    def test_ctor(self):
+        """
+        Refresher object should contain a few handlers
+        :param refresher:
+        :return:
+        """
+        refresher = CompletionRefresher()
+        assert len(refresher.refreshers) > 0
+        actual_handlers = set(refresher.refreshers.keys())
+        expected_handlers = set(['databases', 'schemata', 'tables', 'types', 'views'])
```
pensivebrian: Is 'schemata' a typo?

mssqlcli/mssqlqueries.py
```diff
@@ -90,8 +90,7 @@ def get_views():
     return '''
         SELECT  table_schema,
                 table_name
-        FROM INFORMATION_SCHEMA.TABLES
-        WHERE table_type = 'View'
+        FROM INFORMATION_SCHEMA.VIEWS
```
pensivebrian: My guess as to why this isn't working is you are running against a database with case-sensitive collation, so 'View' doesn't match any rows.

geleems: I see, can we change it to `INFORMATION_SCHEMA.VIEWS` so that we don't need to worry about casing?

tests/test_completion_refresher.py
geleems: No. `schemata` is one of the refresher defined at `completion_refresher.py`

mssqlcli/mssqlqueries.py
pensivebrian: I'm fine with the change to VIEWS

tests/test_completion_refresher.py
geleems: Some of our unit tests are already using unittest lib, and some are using pytest lib. They are different framework that are both supported in python. I could not able to make auto test discovery when using pytest, and it kept popping an error message saying pytest lib is not installed even though it was installed obviously. I found GitHub thread for this bug, and it seems not fixed yet. Class type unit test that extends `unittest.TestCase` (such as `test_mssqlcliclient.py`) with unittest lib are a lot more convenient to run and manage. So I switched to unittest lib as `test_mssqlcliclient.py` does.

pensivebrian: Sounds good. Please make sure the tests still run using the pytest command line.

tests/test_completion_refresher.py
pensivebrian: I still think it's typo. Can we rename this to 'schemas' ?

geleems: Fixed.

PR #260, opened 2019-07-31
Fixed broken unit test test_config.py and test_naive_completion.py

geleems:
Fixed broken unit test test_config.py and test_naive_completion.py

* test test_config.py was using wrong API for creating temporary directory.
* test_naive_completion.py: these tests were using wrong `expected` values and wrong API for creating `set` from `completer.get_completions()` method 
  * test_empty_string_completion
  * test_select_keyword_completion
  * test_function_name_completion
  * test_column_name_completion
  * test_paths_completion
  * test_alter_well_known_keywords_completion

### DIFF
--- a/build.py
+++ b/build.py
@@ -122,7 +122,9 @@ def unit_test():
     utility.exec_command(
         'pytest --cov mssqlcli '
         'tests/test_mssqlcliclient.py '
-        'tests/test_completion_refresher.py ',
+        'tests/test_completion_refresher.py '
+        'tests/test_config.py '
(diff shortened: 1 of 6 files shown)


geleems:
I found there is 1 test failure in `test_naive_completion.py` when I ran with Python 2.7. It turned out there is a bug in `MssqlCompleter.find_matches()` method. When Python 2, the method trys to create Completion object with ASCII string where only unicode string is accepted. All strings in Python 3 are unicode, so it was not blown up on Python 3.

PR #261, opened 2019-08-07
Fixed broken unit tests in 'test_sqlcompletion.py' and made non-class type unit tests into class type

geleems:
* test_json_rpc_contracts.py
  * Renamed `get_test_baseline()` into `get_baseline()` since it was recognized as a unit test case by unit test auto discovery
* tables.py
  * `raise StopIteration` was used previously, which is ok in Python2 but not in Python3. `StopIteration` is converted to `RuntimeError`, and `raise StopIteration` throws exception in Python3. This broke several unit tests in test_sqlcompletion.py running in Python3. Proper way for backward compatibility is using `return` statement. (https://www.python.org/dev/peps/pep-0479/)
* test_prioritization.py and test_rowlimit.py 
  * Converted into class type unit tests
* test_sqlcompletion.py
  * Fixed broken unit tests caused by `StopIteration` RuntimeError in Python3
  * Fixed broken unit test that was disabled: test_sub_select_multiple_col_name_completion
  * Converted into class type unit tests

All tests are passed in both Python2 and 3.

### DIFF
--- a/mssqlcli/packages/parseutils/tables.py
+++ b/mssqlcli/packages/parseutils/tables.py
@@ -43,12 +43,6 @@ def extract_from_part(parsed, stop_at_punctuation=True):
                 # Also 'SELECT * FROM abc JOIN def' will trigger this elif
                 # condition. So we need to ignore the keyword JOIN and its variants
                 # INNER JOIN, FULL OUTER JOIN, etc.
-                
-                # Unlikely in Python 2, StopIteration is converted into a RuntimeError
-                # when code runs 'raise StopIteration' in Python 3.
-                # Generators which explicitly raise StopIteration can generally be
-                # changed to simply return instead.
-                # (https://www.python.org/dev/peps/pep-0479/)
                 return
             elif item.ttype is Keyword and (
                     not item.value.upper() == 'FROM') and (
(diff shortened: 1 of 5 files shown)


mssqlcli/packages/parseutils/tables.py
```diff
@@ -35,15 +35,21 @@ def extract_from_part(parsed, stop_at_punctuation=True):
...
+                # FROM.
+                # Also 'SELECT * FROM abc JOIN def' will trigger this elif
+                # condition. So we need to ignore the keyword JOIN and its variants
+                # INNER JOIN, FULL OUTER JOIN, etc.
+                
+                # Unlikely in Python 2, StopIteration is converted into a RuntimeError
+                # when code runs 'raise StopIteration' in Python 3.
+                # Generators which explicitly raise StopIteration can generally be
+                # changed to simply return instead.
+                # (https://www.python.org/dev/peps/pep-0479/)
```
pensivebrian: Please add these details to the PR description, but I don't think we need these details in the code comments as it's pretty clear what the return does.

geleems: removed.

PR #263, opened 2019-08-08
Fixed broken functionalities when running on Python2

geleems:
Some code is broken when running on Python2. 
* `str` type vs `unicode` type
  * string is `unicode` type in Python3 is , but `str` type in Python2
  * In many places, the updated prompt-toolkit has assertion to check if string param is in `unicode`
* User-input string from command prompt (such as db password) contains `\r` character at the end, which leads to invalid credential failure when connecting to db.

### DIFF
--- a/mssqlcli/main.py
+++ b/mssqlcli/main.py
@@ -63,7 +63,10 @@ def configure_and_update_options(options):
         if not options.username:
             options.username = input(u'Username (press enter for sa):') or u'sa'
         if not options.password:
-            options.password = getpass.getpass()
+            pw = getpass.getpass()
+            if (pw is not None):
+                pw = pw.replace('\r', '').replace('\n', '')
+            options.password = pw
 
 
 def create_config_dir_for_first_use():
--- a/mssqlcli/mssqlqueries.py
+++ b/mssqlcli/mssqlqueries.py
@@ -1,12 +1,16 @@
+import sys
+import re
+
 def get_schemas():
     """
     Query string to retrieve schema names.
     :return: string
     """
-    return '''
+    sql = '''
         SELECT name
         FROM sys.schemas
         ORDER BY 1'''
+    return normalize(sql)
 
 
 def get_databases():
@@ -14,10 +18,11 @@ def get_databases():
     Query string to retrieve all database names.
     :return: string
     """
-    return '''
+    sql = '''
         Select name
         FROM sys.databases
         ORDER BY 1'''
+    return normalize(sql)
 
 
 def get_table_columns():
@@ -25,7 +30,7 @@ def get_table_columns():
     Query string to retrieve all table columns.
     :return: string
     """
-    return '''
+    sql = '''
         SELECT  isc.table_schema,
                 isc.table_name,
                 isc.column_name,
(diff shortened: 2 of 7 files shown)


mssqlcli/mssqlqueries.py
```diff
@@ -164,3 +175,13 @@ def get_foreignkeys():
...
             ORDER BY 3, 4'''
+    return normalize(sql)
+
+
+def normalize(sql):
+    if (sql == '' or sql is None):
+        return sql
+    else:
+        sql = sql.replace('\r', ' ').replace('\n', ' ').strip()
+        sql = re.sub(r' +', ' ', sql)
```
pensivebrian: we didn't neet these replace and substitutions before. Why now?

mssqlcli/mssqlqueries.py
```diff
@@ -80,72 +86,77 @@ def get_view_columns():
...
+    return normalize(sql)
 
 
 def get_views():
     """
     Query string to retrieve all views.
     :return: string
     """
-    return '''
+    sql = '''
```
pensivebrian: We can't add a ` u ` string prefix and assume unicode?

mssqlcli/main.py
```diff
@@ -63,7 +63,10 @@ def configure_and_update_options(options):
         if not options.username:
             options.username = input(u'Username (press enter for sa):') or u'sa'
         if not options.password:
-            options.password = getpass.getpass()
+            pw = getpass.getpass()
```
pensivebrian: why is this breaking now? we ran on python 2/3 with this code as is.

geleems: As I mentioned in the description, when I type `abcd` for password in **command prompt**and hit `ENTER`, `getpass.getpass()` returns `abcd\r`, and it is stored in `options.password`. This made login failed to db since it always have `\r` character for the `ENTER` key in Python2

pensivebrian: But getpass isn't prompt_tookit. Have your tried `pw = getpass.getpass().rstrip()`? Seems more concise and take care of newlines and whitespace.

pensivebrian: nevermind - I like your approach better since a space can be part of a password.

mssqlcli/mssqlqueries.py
geleems: I noticed some of our code recognized the type of command by looking at a word before the first space character. For example:
```
# mssqlcli\packages\special\main.py

def parse_special_command(sql):

    # split the sql by space -- 'command' is a word before first space.
    command, _, arg = sql.partition(' ')

    verbose = '+' in command
    command = command.strip().replace('+', '')
    return (command, verbose, arg.strip())
```

This assumes the word before the first space character is keyword of command (very fragile). Unfortunately we have methods that returns sql with bunch of leading white spaces. For example:
```
def get_sql():
    return '''
        SELECT col1,
               col2
        FROM t1
        '''
```
The method returns a string:
```
'\n        SELECT col1,\n               col2\n        FROM t1\n        '
```
The word before the first space is `\n`, which is recognized as a command keyword. This not intended.
I understand the original author used ''' type string for free-formatting, but I believe these methods have to return correctly formatted such as:
```
'SELECT col1, col2 FROM t1'
```
The `normalize()` method I wrote converts the raw format into correct format.

mssqlcli/mssqlqueries.py
geleems: I can put `u` in front of `'''` for making unicode. But we still need `normalize()` method for removing unintended spaces.

PR #273, opened 2019-08-23
Don't overwrite test results and code coverage

pensivebrian:
Each test results.xml will use a guid in the name, and code coverage will append. This way running tests in tox for 27 won't overwrite the results of 36.


pensivebrian:
@MrMeemus is lurking! How about a CR :)

Issue #277, opened 2019-09-06
Select statement with foreign language fails


geleems:
Select statement with foreign language fails with error message:
```
'ascii' codec can't encode characters in position 0-1: ordinal not in range(128)
```

### Environment:
* Python 2.7
* OS: can repro in any OS

### Repro Steps
1. Create a table of foreign language as following in the target db:
```
create table dbo.한국어테이블
(
    컬럼1 nvarchar(max),
    컬럼2 int
)
insert into dbo.한국어테이블 values(N'테스트1', 1)
```

2. Run select statement in mssql-cli against the target db
```
select * from dbo."한국어테이블"
```

PR #278, opened 2019-09-09
Fix: Select statement with foreign language fails

geleems:
This fix resolves https://github.com/dbcli/mssql-cli/issues/277

### DIFF
--- a/tests/test_mssqlcliclient.py
+++ b/tests/test_mssqlcliclient.py
@@ -210,7 +210,7 @@ class MssqlCliClientTests(unittest.TestCase):
             Verify the column names and string values in rows returned by select statement are properly encoded as unicode.
         """
         local_machine_name = socket.gethostname().replace('-','_').replace('.','_')
-        table_name = 'mssqlcli_test_{0}_{1}'.format(local_machine_name, random_str())
+        table_name = 'mssqlcli_{0}_{1}'.format(local_machine_name, random_str())
         try:
             client = create_mssql_cli_client()
             setup_query = u"CREATE TABLE {0} (컬럼1 nvarchar(MAX), 컬럼2 int);"\
--- a/tests/test_mssqlcliclient.py
+++ b/tests/test_mssqlcliclient.py
@@ -3,6 +3,7 @@ import os
 import io
 import unittest
 import mssqlcli.sqltoolsclient as sqltoolsclient
+from utility import random_str
 
 from mssqlcli.jsonrpc.jsonrpcclient import JsonRpcWriter
 from mssqltestutils import (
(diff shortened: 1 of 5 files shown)


geleems:
@pensivebrian
> Can you add a unit test for this?

Did you mean unit test for `utility.encode()` and `utility.decode()` methods? Or unit test for `mssqlclient._generate_query_results_to_tuples()`?

I can add unit tests for `utility.encode()` and `utility.decode()`, but have no idea how to create unit tests for `mssqlclient._generate_query_results_to_tuples()`, and probably need your help.

pensivebrian:
A unit test that prevents this bug in the future, so `mssqlclient._generate_query_results_to_tuples()`

geleems:
@pensivebrian 
I added unit test that prevents this bug.

tests/test_mssqlcliclient.py
```diff
@@ -202,6 +203,37 @@ def test_stored_proc_multiple_result_sets(self):
...
+    def test_select_result_unicode_encoded(self):
+        """
+            Verify the column names and string values in rows returned by select statement are properly encoded as unicode.
+        """
+        db_name = 'db_' + random_str()
+        table_name = 'table_' + random_str()
+        try:
+            client = create_mssql_cli_client()
+            setup_query = u"CREATE DATABASE {0};"\
+                u"USE {0};"\
```
pensivebrian: This test class assumes an AdventureWorks connection. Can we reuse the existing database?

pensivebrian: >                 [](start = 0, length = 16)

The daily tests run against Azure which don't support this construct. Please run tests against azure.

tests/test_mssqlcliclient.py
```diff
@@ -202,6 +203,37 @@ def test_stored_proc_multiple_result_sets(self):
...
+                u"CREATE TABLE {1} (컬럼1 nvarchar(MAX), 컬럼2 int);"\
+                u"INSERT INTO {1} VALUES (N'테스트1', 1);"\
+                u"INSERT INTO {1} VALUES (N'테스트2', 2);"\
+                .format(db_name, table_name)
```
pensivebrian: Can we run these on a temporary table to assure that table is cleaned up when the connection closes?

tests/test_mssqlcliclient.py
```diff
@@ -202,6 +203,37 @@ def test_stored_proc_multiple_result_sets(self):
         finally:
             shutdown(client)
 
+    
+    def test_select_result_unicode_encoded(self):
+        """
+            Verify the column names and string values in rows returned by select statement are properly encoded as unicode.
+        """
+        db_name = 'db_' + random_str()
```
pensivebrian: >db_ [](start = 19, length = 3)

let's add the local machine name and mssqlcli to the database name. We use a shared server, so this will give folks some idea who is creating this database.

pensivebrian: >db_name [](start = 8, length = 7)

While we're at it, should we test a unicode database name?

pensivebrian:
On a general note, we're having some instability in the nightly runs. If the cloud is having a bad day, a create database statement may fail with a command timeout, but the operation may be in-progress. We should check the server sys.dm_operation_status DMV for the status of the create database operation, and poll until the create database operation completes.

tests/test_mssqlcliclient.py
geleems: Creating a test database is removed from the unit test code.

geleems: USE statement is no longer used in the unit test.

tests/test_mssqlcliclient.py
geleems: I added drop table statement for the clean up.

tests/test_mssqlcliclient.py
geleems: Added machine name and 'mssqlcli' string in the table name as requested.

geleems: Creating a test database is removed from the unit test code. This line no longer exists.

tests/test_mssqlcliclient.py
```diff
@@ -202,6 +204,35 @@ def test_stored_proc_multiple_result_sets(self):
         finally:
             shutdown(client)
 
+    
+    def test_select_result_unicode_encoded(self):
+        """
+            Verify the column names and string values in rows returned by select statement are properly encoded as unicode.
+        """
+        local_machine_name = socket.gethostname().replace('-','_').replace('.','_')
+        table_name = 'mssqlcli_test_{0}_{1}'.format(local_machine_name, random_str())
```
pensivebrian: >test_ [](start = 31, length = 5)

we can remove 'test' from the name as it's implied.

tests/test_mssqlcliclient.py
pensivebrian: If I look on the Adventure works database, I see a bunch of leaked resources from the mssql-cli tests. A temporary table would prevent this.

---
In reply to: [322518262](https://github.com/dbcli/mssql-cli/pull/278#discussion_r322518262) [](ancestors = 322518262)

tests/test_mssqlcliclient.py
geleems: removed as requested.

tests/test_mssqlcliclient.py
pensivebrian: This is still something we need to test. Let's add that in another PR.

tests/test_mssqlcliclient.py
geleems: I have no idea how to convert this sql script into create-temporary-table script.
```
CREATE TABLE table_name (컬럼1 nvarchar(MAX), 컬럼2 int);
INSERT INTO table_name VALUES (N'테스트1', 1);
INSERT INTO table_name VALUES (N'테스트2', 2);
```
Please let me know.

pensivebrian: Just add a `#` prefix: https://docs.microsoft.com/en-us/sql/t-sql/statements/create-table-transact-sql?view=sql-server-2017#temporary-tables

geleems: fixed.

PR #294, opened 2019-09-15
Added .env to .gitignore

ellbosch:
Added .env to .gitignre. Cleared local cache to reset ignored files which removed multiple files in .vscode folder.


ellbosch:
@pensivebrian thoughts on this? Some pros/cons to removing the .env file, we'd probably need to update our build instructions so that developers are told to add their own .env file.

ellbosch:
Removed .vscode from .gitignore so that launch and settings files could be included. However, noticed some issues with changes to the files. Will re-assess and make another commit.

pensivebrian:
Ideally, we'd leave .vscode in the .gitignore so that changes to the launch and settings .json files don't show up in git. At the same time, we check them in so that the files are seeded when you run a git clone (I hope this is how is works). This does mean when we want to update one of these .json files, we have to explicitly force the change via the -f option.

.env
```diff
@@ -1,4 +0,0 @@
-MSSQL_CLI_SERVER=
```
pensivebrian: We still want this file to be checked in so that devs know what variables to set.

.gitignore
```diff
@@ -112,8 +112,8 @@ env*/
 # Pycharm
 .idea/
 
-# vscode
-.vscode/
```
pensivebrian: I don't think we want to delete this.

ellbosch: Right, I'll go fix that. Thanks Brian.

PR #295, opened 2019-09-15
Updated build instructions

ellbosch:
Updated build instructions to include installation of dependencies

### DIFF
--- a/doc/development_guide.md
+++ b/doc/development_guide.md
@@ -113,7 +113,7 @@ General development steps that apply to both a virtual environment or a global e
 
 cd <workspaceRoot>
 python -m pip install --upgrade pip
-python -m pip install -r requirements-dev.txt
+python dev_setup.py
 python build.py build
 ```
 ##### MacOS/Linux (Python3.x)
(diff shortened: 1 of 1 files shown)


doc/development_guide.md
```diff
@@ -113,6 +113,7 @@ General development steps that apply to both a virtual environment or a global e
 
 cd <workspaceRoot>
 python -m pip install --upgrade pip
+python -m pip install -r requirements-dev.txt
```
pensivebrian: this should be `python dev_setup.py`, which runs the above command in addition to some other operations.

pensivebrian: Applies to the other lines as well.

ellbosch: Yes, turns out this should run in our build.py so calling this again shouldn't be necessary. I'll close this PR and not apply the changes.

pensivebrian: No, this does need to be added, otherwise dependencies needed to run build.py won't be installed.

ellbosch: @pensivebrian what about line 55 in the build.py file? It seemed that the goal of this file was to install dependencies as well (although it hasn't worked for me).

pensivebrian: That's for building platform specific wheels when building rpm and deb package for Linux. There is a hook to invoke a custom version of PIP. We probably could document this a lot better. However, I am wondering if that should be requirements.txt (we don't have this but should) and not dev-requirements.txt. Could be a bug there.

ellbosch: Re-opened and made new commit with suggested changes. I'm still running into Mac-related development issues, but this appears to be unrelated.

PR #297, opened 2019-09-18
Fixed broken schema metadata when having unicode in schema name

geleems:
* Fixed broken schema metadata when having unicode in schema name
* Added globalization unit test for schema metadata for intellisense.

### DIFF
--- a/tests/test_globalization.py
+++ b/tests/test_globalization.py
@@ -20,50 +20,50 @@ class GlobalizationResultSetTests(unittest.TestCase):
     
     def test_charset_double(self):
         charset = GB18030()
-        characters_list = (charset.
-            get_next_characters(charset.get_double_characters()))
+        characters_list = charset.get_next_characters(
+            charset.get_double_characters())
         self.run_charset_validation(characters_list)
     
 
     def test_charset_four(self):
         charset = GB18030()
-        characters_list = (charset.
-            get_next_characters(charset.get_four_characters()))
+        characters_list = charset.get_next_characters(
+            charset.get_four_characters())
         self.run_charset_validation(characters_list)
 
 
     def test_charset_mongolian(self):
         charset = GB18030()
-        characters_list = (charset.
-            get_next_characters(charset.get_mongolian_characters()))
+        characters_list = charset.get_next_characters(
+            charset.get_mongolian_characters())
         self.run_charset_validation(characters_list)
 
 
     def test_charset_uyghur(self):
         charset = GB18030()
-        characters_list = (charset.
-            get_next_characters(charset.get_uyghur_characters()))
+        characters_list = charset.get_next_characters(
+            charset.get_uyghur_characters())
         self.run_charset_validation(characters_list)
 
 
     def test_charset_tibetian(self):
         charset = GB18030()
-        characters_list = (charset.
-            get_next_characters(charset.get_tibetian_characters()))
+        characters_list = charset.get_next_characters(
+            charset.get_tibetian_characters())
         self.run_charset_validation(characters_list)
 
 
     def test_charset_yi(self):
         charset = GB18030()
-        characters_list = (charset.
-            get_next_characters(charset.get_yi_characters()))
+        characters_list = charset.get_next_characters(
+            charset.get_yi_characters())
         self.run_charset_validation(characters_list)
 
 
     def test_charset_zang(self):
         charset = GB18030()
-        characters_list = (charset.
-            get_next_characters(charset.get_zang_characters()))
+        characters_list = charset.get_next_characters(
+            charset.get_zang_characters())
         self.run_charset_validation(characters_list)
 
     
@@ -108,8 +108,8 @@ class GlobalizationMetadataTests(unittest.TestCase):
 
     def test_schema_metadata_double(self):
         charset = GB18030()
-        characters_list = (charset.
-            get_next_characters(charset.get_double_characters()))
+        characters_list = charset.get_next_characters(
+            charset.get_double_characters())
         self.run_schema_metadata_validation(characters_list)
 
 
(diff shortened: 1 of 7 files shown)


tests/test_globalization.py
```diff
@@ -3,91 +3,148 @@
...
-# All tests require a live connection to test server.
-# Make modifications to mssqlutils.create_mssql_cli_client() to use a different server and database.
-# Please Note: These tests cannot be run offline.
 class GlobalizationResultSetTests(unittest.TestCase):
     
     def test_charset_double(self):
-        charset_list = self.get_charset_double() 
-        self.run_charset_validation(charset_list)
+        charset = GB18030()
+        characters_list = charset.\
```
pensivebrian: nit: rather than use explicit line continuations via the backslash, can you break on a { or a ( which are implicit line continuations?

utility.py
```diff
@@ -83,17 +83,21 @@ def copy_current_platform_mssqltoolsservice():
 
 
 def encode(s):
-    if (s is not None and sys.version_info.major < 3):
+    try:
```
pensivebrian: Why the change to try catch? Should we really catch all exceptions?

tests/test_globalization.py
```diff
@@ -3,91 +3,148 @@
...
+                    assert is_error is False
+                    new_schemas.append(schema_name)
+            completion_refresher.refresh_schemas(completer, client)
+            completions = completer.\
+                get_completions(Document(u'select * from ', 14, None), None, True)
+            db_schemas = set(map(lambda e: e.text, completions))
+            for new_schema in new_schemas:
+                assert u'"{}"'.format(new_schema) in db_schemas
+        finally:
+            for new_schema in new_schemas:
```
pensivebrian: Should we wrap this in a try catch so that any cleanup exception won't overwrite an assert exception?

tests/test_globalization.py
```diff
@@ -3,91 +3,148 @@
...
+            get_next_characters(charset.get_double_characters())
+        self.run_schema_metadata_validation(characters_list)
+
+
+    def run_schema_metadata_validation(self, characters_list):
+        client = create_mssql_cli_client()
+        completer = MssqlCompleter(smart_completion=True)
+        new_schemas = []
+        try:
+            local_machine_name = socket.gethostname().\
```
pensivebrian: We have machine_name logic in two places now. Move into a common function?

tests/test_globalization.py
```diff
@@ -3,91 +3,148 @@
...
+            for characters in characters_list:
+                schema_name = u'mssqlcli_{0}_{1}_{2}'.\
+                    format(local_machine_name, random_str(), characters)
+                query = u'CREATE SCHEMA {0}'.format(schema_name)
+                for _, _, _, _, is_error in client.execute_query(query):
+                    assert is_error is False
+                    new_schemas.append(schema_name)
+            completion_refresher.refresh_schemas(completer, client)
+            completions = completer.\
+                get_completions(Document(u'select * from ', 14, None), None, True)
```
pensivebrian: nit: named arguments for readability

tests/test_globalization.py
```diff
@@ -3,91 +3,148 @@
...
+        completer = MssqlCompleter(smart_completion=True)
+        new_schemas = []
+        try:
+            local_machine_name = socket.gethostname().\
+                replace('-','_').replace('.','_')
+            for characters in characters_list:
+                schema_name = u'mssqlcli_{0}_{1}_{2}'.\
+                    format(local_machine_name, random_str(), characters)
+                query = u'CREATE SCHEMA {0}'.format(schema_name)
+                for _, _, _, _, is_error in client.execute_query(query):
```
pensivebrian: Do we need a for loop?

tests/test_globalization.py
```diff
@@ -3,91 +3,148 @@
...
+                test_str = characters
+                col1_name = u'col_{0}1'.format(test_str)
+                col2_name = u'col_{0}2'.format(test_str)
+                table_name = u'#mssqlcli_{0}_{1}_{2}'.format(local_machine_name, random_str(), test_str)
+                setup_query = u"CREATE TABLE {0} ({1} nvarchar(MAX), {2} int);"\
+                    u"INSERT INTO {0} VALUES (N'value_{3}1', 1);"\
+                    u"INSERT INTO {0} VALUES (N'value_{3}2', 2);"\
+                    .format(table_name, col1_name, col2_name, test_str)
+                
+                for rows, columns, status, statement, is_error in client.execute_query(setup_query):
```
pensivebrian: For loop needed?

utility.py
geleems: There are several cases that it throws exception which should be ignored.
* when it is python3
* when argument is not string type
* when argument is already encoded/decoded

tests/test_globalization.py
geleems: fixed

tests/test_globalization.py
geleems: fixed as you requested. I am very sure it will complain on the implicit way when the project uses pep8

tests/test_globalization.py
geleems: fixed.

tests/test_globalization.py
geleems: fixed.

tests/test_globalization.py
geleems: unfortunately yes. client.execute_query(query) is a generator.

tests/test_globalization.py
geleems: unfortunately yes. client.execute_query(query) is a generator.

pensivebrian: dang. we need to add execute_non_query and execute_scalar methods some day.

geleems: yup :)

tests/test_globalization.py
```diff
@@ -1,93 +1,156 @@
...
-# Make modifications to mssqlutils.create_mssql_cli_client() to use a different server and database.
-# Please Note: These tests cannot be run offline.
 class GlobalizationResultSetTests(unittest.TestCase):
     
     def test_charset_double(self):
-        charset_list = self.get_charset_double() 
-        self.run_charset_validation(charset_list)
+        charset = GB18030()
+        characters_list = (charset.
+            get_next_characters(charset.get_double_characters()))
```
pensivebrian: I thought implicit line continuation are for parentheses, brackets and bracess. For this case I think the open paren seems more natural for readability.

pensivebrian: For example:

```
characters_list = charset.get_next_characters(
    charset.get_double_characters())
```

geleems: fixed as requested.

PR #300, opened 2019-09-20
Added unit test for table metadata globalization

geleems:
Added unit test for table metadata globalization

### DIFF
--- a/tests/test_globalization.py
+++ b/tests/test_globalization.py
@@ -1,10 +1,14 @@
 # coding=utf-8
 import unittest
 import socket
-import os
 from utility import random_str
-from mssqltestutils import (create_mssql_cli_client,
-    create_mssql_cli_options, shutdown)
+from mssqltestutils import (
+    create_mssql_cli_client,
+    create_mssql_cli_options,
+    create_test_db,
+    clean_up_test_db,
+    shutdown
+)
 from mssqlcli.mssqlcompleter import MssqlCompleter
 import mssqlcli.completion_refresher as completion_refresher
 from prompt_toolkit.document import Document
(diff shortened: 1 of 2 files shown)


tests/test_globalization.py
```diff
@@ -106,47 +111,107 @@ def run_charset_validation(self, characters_list):
...
-
-
-    def run_schema_metadata_validation(self, characters_list):
-        client = create_mssql_cli_client()
-        completer = MssqlCompleter(smart_completion=True)
-        new_schemas = []
+    def test_schema_metadata(self):
+        characters_list = get_GB18030_characters_list()
         try:
+            client1 = create_mssql_cli_client()
```
pensivebrian: Can you move the create database operation into the mssqltestutils.py so that database creation is done in one place? Ideally, it's returns a object with a Drop method and Name property, but we can start with just returning the name.

tests/test_globalization.py
```diff
@@ -232,4 +297,16 @@ def get_next_characters(self, all_characters, count=50):
             next_characters_length = min(len(all_characters), count)
             next_characters = all_characters[0:next_characters_length]
             all_characters = all_characters[next_characters_length:]
-            yield next_characters
\ No newline at end of file
+            yield next_characters
+    
+    def get_all_characters(self):
```
pensivebrian: Not a fan of this change since when something fails, we don't know which charset is the problem. Can you add some logging of the charset name as we're testing? Or, just create a test per charset like before?

tests/test_globalization.py
geleems: fixed as requested.

tests/test_globalization.py
geleems: fixed to test per charset like before.

PR #304, opened 2019-10-01
Fixed incorrect pythonpath reference

ellbosch:
The setting for PYTHONPATH was incorrect in launch.json, which led to unfound modules in debug mode. This fix enables debug mode in VS Code to work again.


ellbosch:
Found more issues, closing for now.

PR #309, opened 2019-10-02
Non-interactive query mode

### LINKED ISSUE #242: Execute sql command on cli like -Q in sqlcmd
Is there any way to execute a sql command directly from cli?

Like -Q option in sqlcmd

It would be nice to have something like `mssql-cli -Q 'SELECT * FROM table'` with output on stdout

ellbosch:
This PR would close #242. Unit tests have not been covered yet. Thoughts on the implementation @pensivebrian?


.vscode/launch.json
```diff
@@ -24,6 +24,17 @@
             "program": "${file}",
             "envFile": "${workspaceRoot}/.env",
             "console": "externalTerminal",
-        }
+        },
+        {
+            "name": "Test Query (external)",
```
pensivebrian: Let's remove the lauch.json changes. We should have pytests that launch mssql-cli in non-interactive mode, and validate the output.

mssqlcli/mssql_cli.py
```diff
@@ -371,6 +375,11 @@ def run(self):
             print('Mail: sqlcli@microsoft.com')
             print('Home: http://github.com/dbcli/mssql-cli')
 
+        # if not in interactive mode, don't build CLI
+        if not self.interactive_mode:
```
pensivebrian: Does this by-pass interactive features, such as the pager?

mssqlcli/mssqlclioptionsparser.py
```diff
@@ -149,4 +149,12 @@ def create_parser():
         metavar=u'',
         help=u'Prompt format (Default: \\d> ')
 
+    args_parser.add_argument(
+        u'-Q', u'--query',
+        dest=u'query',
+        metavar=u'',
+        default=False,
+        help=u'Executes a query when mssql-cli starts and immediately exits. Use quotation marks around query.'
```
pensivebrian: "Executes a query outputting the results to stdout and exits."

.vscode/launch.json
ellbosch: We should definitely have pytests—however the usefulness with launch.json is purely for debugging purposes so that we can use breakpoints. Is it possible to also use pytest in this capacity?

mssqlcli/mssql_cli.py
ellbosch: That's the intent—it runs `execute_command` and exits.

.vscode/launch.json
pensivebrian: I see. It would be nice if we can step through the pytest hosted mssql-cli process as well, though I'm not sure what hoops we need to jump through to enable that.

mssqlcli/mssql_cli.py
pensivebrian: Yes, but IIRC, execute_command pipe the results through a pager, which expects interactive key presses to page the results. I would expect us to have to add some logic around this. If we test with some larger output we can confirm the behavior.

ellbosch: I see—I'll look into this.

ellbosch:
Considering possible test cases for -Q:
- Session closes for valid query
- Session closes for invalid query
- Valid query returns same output as `execute_query` output
- Invalid query returns same output as `execute_query` output

Any others @pensivebrian?

mssqlcli/mssql_cli.py
```diff
@@ -302,7 +306,7 @@ def handle_editor_command(self, text):
             editor_command = special.editor_command(text)
         return text
 
-    def execute_command(self, text, query):
+    def execute_command(self, text, query, test_format=False):
```
ellbosch: @pensivebrian pointing this change out for your awareness—this optional param helps a lot with testing but I'm very open to alternative methods you know would work that don't require extensive changes to `execute_command`.

ellbosch:
Latest changes introduced a fix for a bug that caused crashes on exiting VIM mode and more tests. As a disclaimer—I've noticed that calling `mssql-cli -Q ""` with empty strings will launch the default interactive mode. We should have further discussions about expected functionality for empty quotes.

tests/test_noninteractive_mode.py
```diff
@@ -0,0 +1,80 @@
...
+
+def test_session_closure_query_invalid():
+    """ Test session closure for invalid query. """
+    assert is_processed_closed("asdfasoifjas")
+
+def test_query_large():
+    query_str = "SELECT * FROM STRING_SPLIT(REPLICATE(CAST('X,' AS VARCHAR(MAX)), 1024), ',')"
+    p = subprocess.Popen("./mssql-cli -Q \"%s\"" % query_str, shell=True, stdin=subprocess.PIPE,
+                         stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+    output = p.communicate(b'y\n')[0].decode("utf-8")
```
pensivebrian: > output = p.communicate(b'y\n')[0].decode("utf-8") [](start = 3, length = 50)

Doesn't this mean we're interactive?  If that's the case, we should make change to the product code so that we are non-interactive.

ellbosch: This is only when the query exceeds a specified row limit—it prompts the user whether we should load everything. Is this considered interactive mode? If yes, is expected functionality to display everything regardless of query size?

pensivebrian: > This is only when the query exceeds a specified row limit—it prompts the user whether we should load everything. Is this considered interactive mode? If yes, is expected functionality to display everything regardless of query size?

yes, any prompt is considered interactive mode. You bring up a good point, what should the behavior be? We'll need to figure this out.

ellbosch: I suppose we should display the VIM mode with no prompt—exact same functionality but we remove the prompt that shows up if rows exceed row limit. If you agree, I'll send out an update for this.

pensivebrian: I'd see what sqlcmd does. I'm not sure row limit makes sense in non-interactive mode.

ellbosch: Just checked—SQLCMD outputs results without a prompt.

mssqlcli/main.py
```diff
@@ -52,7 +52,12 @@ def run_cli_with(options):
     mssqlcli.connect_to_database()
 
     telemetry_session.set_server_information(mssqlcli.mssqlcliclient_main)
-    mssqlcli.run()
+
+    # don't run cli if query is specified
+    if not options.query:
+        mssqlcli.run()
+    else:
+        mssqlcli.execute_query(str(options.query))
```
pensivebrian: shouldn't we call mssqlcli.shutdown() after we execute the query?

mssqlcli/mssqlcliclient.py
```diff
@@ -50,6 +50,9 @@ def __init__(self, mssqlcli_options, sql_tools_client, owner_uri=None, **kwargs)
         self.server_edition = None
         self.is_cloud = False
 
+        # non-interactive mode options
+        self.query = mssqlcli_options.query
+
```
pensivebrian: Can we remove this now that we are passing in the query string via the execute_query method argument?

tests/test_noninteractive_mode.py
```diff
@@ -0,0 +1,49 @@
...
+                         stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+    output = p.communicate()[0].decode("utf-8").rstrip('\n')
+    output_baseline = get_file_contents('./tests/query_tests/small_query.txt')
+    assert output == output_baseline
+
+def is_processed_closed(query_str):
+    """ Runs unit tests on process closure given a query string """
+    mssql_cli = create_mssql_cli()
+    mssql_cli.execute_query(query_str)
+    mssql_cli.shutdown()
```
pensivebrian: I wouldn't expect this method to execute a query given it's name. Consider renaming?

pensivebrian: Can we move the shutdown call into a finally block so that it's always executed - we don't want to leak processes.

tests/test_noninteractive_mode.py
```diff
@@ -0,0 +1,49 @@
...
+                         stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+    output = p.communicate()[0].decode("utf-8")
+    output_baseline = get_file_contents('./tests/query_tests/big_query.txt')
+    assert output == output_baseline
+
+def test_query_output():
+    """ Tests if -Q has equal output to execute_query. """
+    query_str = "select 1"
+    p = subprocess.Popen("./mssql-cli -Q '%s'" % query_str, shell=True, stdin=subprocess.PIPE,
+                         stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
```
pensivebrian: Rather than duplicate this, can we move it into a factory method that tests call to Open the process?

tests/test_noninteractive_mode.py
```diff
@@ -0,0 +1,49 @@
...
+
+def test_session_closure_query_valid():
+    """ Test session closure for valid query. """
+    assert is_processed_closed("select 1")
+
+def test_session_closure_query_invalid():
+    """ Test session closure for invalid query. """
+    assert is_processed_closed("asdfasoifjas")
+
+def test_query_large():
```
pensivebrian: >test_query_large [](start = 4, length = 16)

I feel like you have two sets of tests: ones that use MssqlCli and ones that execute the process. Can we break these up and put them into classes the leverage the test setup and teardown methods to assert pre/post condition? For example, a common post condition in the MssqlCli client test is to assert the sqltoolsservice process is closed, which could be in the teardown.

pensivebrian:
> 

things like this don't make sense in non-interactive mode.

---
Refers to: mssqlcli/mssql_cli.py:184 in 3700608. [](commit_id = 370060894b4248e957d3045062a47c04d4a98eb4, deletion_comment = False)

mssqlcli/mssql_cli.py
```diff
@@ -353,8 +340,36 @@ def execute_command(self, text, query):
...
                 self.refresh_completions(persist_priorities='all')
 
+        if not query.contains_secure_statement:
+            # Allow MssqlCompleter to learn user's preferred keywords, etc.
+            with self._completer_lock:
+                self.completer.extend_query_history(text)
+
+            self.query_history.append(query)
+
+    def execute_query(self, text, is_interactive=False):
```
pensivebrian: >is_interactive [](start = 34, length = 14)

I'm thinking we should make this an instance member that can be passed in via the options in the constructor. Then you don't have two duplicate methods. Defaulting to True of course.

tests/test_noninteractive_mode.py
ellbosch: @pensivebrian this is exactly what we're testing for—if processes are still running then the assert statement will catch that. So far `shutdown()` has always worked. So I don't think we need to add that, or even _should_ add that so that we can test to see if the client is still running.

ellbosch: Please feel free to leave suggestions :) what do you think is better than `execute_query()`?

tests/test_noninteractive_mode.py
ellbosch: Oh I forgot to deploy that change! Already done.

tests/test_noninteractive_mode.py
ellbosch: Good idea!

mssqlcli/mssql_cli.py
ellbosch: Let's sync on this—I explored this thoroughly and it's a bit complicated.

tests/test_noninteractive_mode.py
```diff
@@ -0,0 +1,92 @@
...
+        assert output_query == output_file
+
+    def test_col_too_wide(self):
+        """ Tests query where column is too wide. """
+        s = 'x' * 250
+        query_str = "select %s" % s
+        file_baseline = self.file_root + 'col_too_wide.txt'
+        output_query, output_file = self.is_query_valid(query_str, file_baseline)
+        assert output_query == output_file
+
```
udeeshagautam: Thinking of test cases do we need ..? 
1. Valid query but no result
2. Cancellation while query is running

(not necessarily for this PR)

ellbosch: Good suggestions, thank you!

mssqlcli/main.py
```diff
@@ -52,19 +45,26 @@ def run_cli_with(options):
...
 
     telemetry_session.set_server_information(mssqlcli.mssqlcliclient_main)
-    mssqlcli.run()
+
+    # don't run cli if query is specified
+    if not options.query:
+        mssqlcli.run()
+    else:
+        mssqlcli.execute_query(str(options.query))
+        mssqlcli.shutdown()
```
pensivebrian: probably should wrap this in a try finally block so that we always call shutdown

mssqlcli/mssql_cli.py
```diff
@@ -150,8 +150,12 @@ def __init__(self, options):
         self.cli_style = c['colors']
         self.output_style = style_factory_output(self.syntax_style, self.cli_style)
         self.wider_completion_menu = c['main'].as_bool('wider_completion_menu')
+
+        # override interactive mode to false if query is set
+        self.interactive_mode = options.interactive_mode and not options.query
```
pensivebrian: I don't think we should have the `and not options.query ` as part of the condition - we should just accept what's passed.

pensivebrian:
> 

Skip in non-interactive mode?

---
Refers to: mssqlcli/mssql_cli.py:188 in fba47c5. [](commit_id = fba47c52b361544dd32efed4edf292d40b405912, deletion_comment = False)

mssqlcli/mssqlclioptionsparser.py
```diff
@@ -149,4 +157,11 @@ def create_parser():
         metavar=u'',
         help=u'Prompt format (Default: \\d> ')
 
+    args_parser.add_argument(
+        u'--interactive_mode',
+        dest=u'interactive_mode',
+        metavar=u'',
+        default=True
+    )
```
pensivebrian: We can remove this since it's inferred based on the arguments passed.

mssqlcli/mssql_cli.py
```diff
@@ -498,7 +524,7 @@ def _evaluate_command(self, text):
                 self.mssqlcliclient_main.execute_query(text):
 
             total = time() - start
-            if self._should_show_limit_prompt(status, rows):
+            if self._should_show_limit_prompt(status, rows) and self.interactive_mode:
                 click.secho('The result set has more than %s rows.'
```
pensivebrian: Can we move the self.interactive_mode check into the _should_show_limit_prompt method? This way, any caller in non-interactive mode will get false.

mssqlcli/mssql_cli.py
```diff
@@ -353,9 +344,54 @@ def execute_command(self, text, query):
...
+
+            self.query_history.append(query)
+
+        return output, query
+
+    def execute_query(self, text):
+        """ Processes a query string and outputs to file or terminal """
+
+        # Initialize default metaquery in case execution fails
+        query = MetaQuery(query=text, successful=False)
```
pensivebrian: We aren't doing anything in this method with the query. Can we move it's instantiation into the _execute_interactive_command method so that it doesn't have to be passed in via a parameter?

ellbosch:
> > 
> 
> Skip in non-interactive mode?
> 
> Refers to: mssqlcli/mssql_cli.py:188 in [fba47c5](https://github.com/dbcli/mssql-cli/commit/fba47c52b361544dd32efed4edf292d40b405912). [](commit_id = [fba47c5](https://github.com/dbcli/mssql-cli/commit/fba47c52b361544dd32efed4edf292d40b405912), deletion_comment = False)

GitHub is having problems showing what line you're referring to—I went to the commit and line # but wasn't able to determine what you're talking about.

mssqlcli/mssqlclioptionsparser.py
ellbosch: Thanks, good call!

tests/test_noninteractive_mode.py
```diff
@@ -0,0 +1,96 @@
...
+    create_mssql_cli,
+    # create_test_db,
+    # clean_up_test_db,
+)
+
+# TODO: split up into multiple classes and create automated way to generate query files
+class TestNonInteractive:
+    """ Tests non-interactive features. """
+
+    file_root = './tests/test_query_results/'
```
pensivebrian: I don't think you can assume root to be the current directory. I think there a python API to get the current file path, from which you can infer the path.

tests/test_noninteractive_mode.py
```diff
@@ -0,0 +1,96 @@
...
+        assert self.is_processed_closed("select 1")
+
+    def test_session_closure_query_invalid(self):
+        """ Test session closure for invalid query. """
+        assert self.is_processed_closed("asdfasoifjas")
+
+    def test_query_large(self):
+        """ Test against query with over 1000 lines. """
+        query_str = "SELECT * FROM STRING_SPLIT(REPLICATE(CAST('X,' AS VARCHAR(MAX)), 1024), ',')"
+        file_baseline = self.file_root + 'big.txt'
```
pensivebrian: A neat convention would be to name the baseline file the same as the test name.

tests/test_noninteractive_mode.py
```diff
@@ -0,0 +1,96 @@
...
+        output_query, output_file = self.get_query_results(query_str, file_baseline)
+        assert output_query == output_file
+
+    # TODO: test that calling run with interactive_mode off returns error
+    def test_invalid_run(self):
+        pass
+
+    # HELPER FUNCTIONS
+
+    def get_query_results(self, query_str, file_baseline):
```
pensivebrian: nit: method name. How about execute_query_via_subprocess?

tests/test_noninteractive_mode.py
```diff
@@ -0,0 +1,96 @@
...
+        """ Helper method for running a query with -Q. """
+        p = subprocess.Popen("./mssql-cli -Q \"%s\"" % query_str, shell=True, stdin=subprocess.PIPE,
+                             stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+        output = p.communicate()[0].decode("utf-8")
+        output_baseline = self.get_file_contents(file_baseline)
+        return output, output_baseline
```
pensivebrian: I wasn't expecting the baseline to be loaded here. I would move loading the baseline into another method.

tests/test_noninteractive_mode.py
pensivebrian: But what if execute_query throws? We still want to shutdown and not raise an assert in this situation since it's expected to be still running.

tests/test_noninteractive_mode.py
ellbosch: I actually decided to use the parametrize pattern in pytest instead, which I validated prints useful error messages in the event of a failure.

tests/test_noninteractive_mode.py
```diff
@@ -0,0 +1,96 @@
...
+    def get_file_contents(file_path):
+        """ Get expected result from file. """
+        try:
+            with open(file_path, 'r') as f:
+                return f.read()
+        except OSError as e:
+            raise e
+
+
+class TestNonInteractiveClosure:
```
pensivebrian: naming: can TestNonInteractiveClosure be more clear? A closure means something very specific to programmers. Maybe TestNonInteractiveShutdown?

tests/test_noninteractive_mode.py
```diff
@@ -0,0 +1,97 @@
...
+        "select 1",
+        "gibberish"
+    ]
+
+    @classmethod
+    def setup_class(cls):
+        """
+        Instantiate empty mssql_cli
+        """
+        cls.mssql_cli = None
```
pensivebrian: Is this used? Remove?

tests/test_noninteractive_mode.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def setup_method(self):
+        """ Create new mssql-cli instance for each test """
+        self.mssql_cli = create_mssql_cli()
+
+    @pytest.mark.parametrize("query_str", testdata)
+    def test_session_closure(self, query_str):
+        """ Runs unit tests on process closure given a query string. """
```
pensivebrian: same naming issue: test_session_shutdown. Also, it's not quite clear what a session is.

tests/test_noninteractive_mode.py
ellbosch: > Some minor comments. The only product issue, it in the MssqlCli constructor, we shouldn't be spinning up the completion refreshers or anything that doesn't make sense in non-interactive mode.

Ok, will give this a shot.

ellbosch:
> Some minor comments. The only product issue, it in the MssqlCli constructor, we shouldn't be spinning up the completion refreshers or anything that doesn't make sense in non-interactive mode.

Ok, will give this a shot.

ellbosch:
@pensivebrian very curious about your thoughts with [this commit](https://github.com/dbcli/mssql-cli/pull/309/commits/bd06d592f01ad906df6a00e33a78fd237e956dc7), which will only set various properties in `MssqlCli` class if interactive mode is set. Tests have passed with these changes but I want to be sure that there aren't any unintended consequences.

pensivebrian:
Looks good. I’d still take the tool for a good test spin manually, querying and changing database to make sure it works.

tests/test_noninteractive_mode.py
```diff
@@ -32,6 +32,19 @@ def test_query(self, query_str, file_baseline):
         output_baseline = self.get_file_contents(file_baseline)
         assert output_query == output_baseline
 
+    @staticmethod
+    def test_noninteractive_run():
+        """ Test that calling run throws an exception only when interactive_mode is false """
+        mssqlcli = create_mssql_cli(interactive_mode=False)
+        try:
+            mssqlcli.run()
+            mssqlcli.shutdown()
```
ellbosch: @pensivebrian I had some problems getting `shutdown()` to call in the event that `mssqlcli` actually spins up interactive mode. Is there a way to shutdown `mssqlcli` from the code?

pensivebrian: Run throws an exception which skips shutdown. Put run in a try statement block and shutdown in a finally statement block.

tests/test_noninteractive_mode.py
```diff
@@ -63,6 +63,47 @@ def get_file_contents(file_path):
...
+        yield test_db
+        clean_up_test_db(test_db)
+
+    @staticmethod
+    @pytest.fixture(scope="function")
+    def mssqlcli(test_db):
+        return create_mssql_cli(interactive_mode=False, database=test_db)
+
+    """
+    TODO: TESTS TO CONSIDER:
```
pensivebrian: All of these scenarios would take the same code path, no? What value does each scenario bring?

ellbosch: Good point. It's the same code path, so perhaps no added value, at least for testing non-interactive mode. Any you can think of that may add value?

mssqlcli/main.py
```diff
@@ -48,23 +41,36 @@ def run_cli_with(options):
...
 
-    telemetry_session.set_server_information(mssqlcli.mssqlcliclient_main)
-    mssqlcli.run()
+    mssqlcli = MssqlCli(options)
+    try:
+        mssqlcli.connect_to_database()
+        telemetry_session.set_server_information(mssqlcli.mssqlcliclient_main)
+
+        if mssqlcli.interactive_mode:
+            # run cli if in interactive mode
```
pensivebrian: we can remove these code comments as the code is already explicit as to whats going on and the comments aren't adding any value.

tests/test_noninteractive_mode.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+class TestNonInteractiveDbModification:
+    """
+    Test DB modification in non-interactive mode.
+    """
+
+    @staticmethod
+    @pytest.fixture(scope="class")
+    def test_db():
```
pensivebrian: Let's just remove the TestNonInteractiveDbModification test - they repeat the TestNonInteractiveShutdown code path.

ellbosch:
My latest changes fix all known python 2 issues. So far all tests have passed, I've also successfully created builds remotely on Azure. I'm running one more just to be sure.

tests/test_noninteractive_mode.py
```diff
@@ -28,6 +26,48 @@ def test_query(self, query_str, file_baseline):
         output_baseline = self.get_file_contents(file_baseline)
         assert output_query == output_baseline
 
+    def test_long_query(self):
+        """ Special test that inputs a SQLCMD function before outputting a large query.
+        Needed to circumvent compat issues with AdventureWorks2014 db which doesn't
+        support SPLIT_STRING. """
+
+        split_string_func = """CREATE FUNCTION dbo.splitstring ( @stringToSplit VARCHAR(MAX) )\n
```
pensivebrian: Can't we just run against a 130 compat level or higher? I prefer not workaround this by creating a function, especially it it's a shared/read only database like AdventureWorks.

ellbosch: My bad, I completely forgot to make this change. On it now.

mssqlcli/sqltoolsclient.py
```diff
@@ -90,15 +90,22 @@ def shutdown(self):
...
+                self.tools_service_process.kill()
+                time.sleep(0.1)
+            except OSError:
+                # catches 'No such process' error from Python 2
+                pass
+            finally:
+                # Close the stdout file handle or else we would get a resource warning (found via
+                # pytest). This must be closed after the process is killed, otherwise we would block
+                # because the process is using it's stdout.
+                self.tools_service_process.stdout.close()
```
pensivebrian: > self.tools_service_process.stdout.close() [](start = 15, length = 42)

Shouldn't this be in the try? I would guess that this is the

PR #318, opened 2019-10-23
windows file path support

ellbosch:
Quick fix for Windows support for the -Q PR.

### DIFF
--- a/tests/test_noninteractive_mode.py
+++ b/tests/test_noninteractive_mode.py
@@ -6,7 +6,9 @@ import os
 import pytest
 from mssqltestutils import create_mssql_cli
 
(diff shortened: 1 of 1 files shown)


tests/test_noninteractive_mode.py
```diff
@@ -6,7 +6,9 @@
 import pytest
 from mssqltestutils import create_mssql_cli
 
-_FILE_ROOT = '%s/test_query_results/' % os.path.dirname(os.path.abspath(__file__))
+_SLASH = '\\' if os.name == 'nt' else '/'
```
pensivebrian: I think you can use os.path.join() to create file paths without having use the underlying platform specific path separator.

PR #325, opened 2019-10-29
fixed reference to file path for official build downloads

ellbosch:
Fixed an issue where official builds would download to the wrong location.

### DIFF
--- a/release.py
+++ b/release.py
@@ -90,9 +90,8 @@ def download_official_wheels():
             print('Error: blob: {} does not exist in container: {}'.format(blob, BLOB_MSSQL_CLI_DAILY_CONTAINER_NAME))
             sys.exit(1)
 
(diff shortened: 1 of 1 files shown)


release.py
```diff
@@ -90,9 +90,8 @@ def download_official_wheels():
             print('Error: blob: {} does not exist in container: {}'.format(blob, BLOB_MSSQL_CLI_DAILY_CONTAINER_NAME))
             sys.exit(1)
 
-        blob_service.get_blob_to_path(BLOB_MSSQL_CLI_DAILY_CONTAINER_NAME, blob, '{0}\{1}'.format(
-                                                                                        utility.MSSQLCLI_DIST_DIRECTORY,
-                                                                                        blob))
+        blob_service.get_blob_to_path(BLOB_MSSQL_CLI_DAILY_CONTAINER_NAME, blob,
```
udeeshagautam: what will be the output format of  os.path.join? it will change the \ / based on OS?

PR #326, opened 2019-10-30
non-interactive support for input file with -i

### LINKED ISSUE #305: Feature Request: Support -i to execute a .sql file
This is important for sqlcmd parity.

First priority is non-interactive support.
Second is interactive support. special command through -i

https://docs.microsoft.com/sql/tools/sqlcmd-utility?redirectedfrom=MSDN&view=sql-server-2017

ellbosch:
Closes #305


mssqlcli/mssql_cli.py
```diff
@@ -366,6 +368,20 @@ def execute_query(self, text):
             output = self._execute_interactive_command(text)
         else:
             # non-interactive mode
+            if self.input_file:
+                if text:
+                    # return error if both query text and an input file are specified
+                    click.secho("Invalid arguments: either -Q or -i may be specified.",
```
kevcunnane: Minor: a common pattern is to validate fields at the start (e.g. -Q vs -i) and then have streamlined logic. This can make it cleaner as additional requirements come up later. Not required, just something to think about

kevcunnane: Should you use the full names of these arguments in the error message? What if someone doesn't know what -i or -Q means?

mssqlcli/mssql_cli.py
```diff
@@ -366,6 +368,20 @@ def execute_query(self, text):
             output = self._execute_interactive_command(text)
```
kevcunnane: I'm confused. You no longer call str(text) anywhere. Why is it not needed? Given removal in outer class I'm surprised we don't need to do anything here?

mssqlcli/main.py
```diff
@@ -53,7 +53,7 @@ def run_cli_with(options):
         if mssqlcli.interactive_mode:
             mssqlcli.run()
         else:
-            mssqlcli.execute_query(str(options.query))
+            mssqlcli.execute_query(options.query)
```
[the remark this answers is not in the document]
ellbosch: Tests were passing without the string conversion so I decided to omit it. I'd be glad to add it back if you think it's needed.

You're right on the second bullet. Logic is not consistent for input files so I'll rework this.

mssqlcli/mssql_cli.py
ellbosch: Thanks @kevcunnane, agreed. I'll look into moving functionality to constructor.

mssqlcli/mssql_cli.py
ellbosch: I'm unsure how `text` could ever end up not having a string type, but to be sure I can throw in a conversion at the beginning of the function call for `execute_query`.

PR #327, opened 2019-10-31
using random strings for entities in test

### LINKED ISSUE #281: Test stability
We have 2-3 test run failures a week with our dailly runs. Tests need to be more robust (retry logic?) and use unique names so they don't step on each other.

ellbosch:
This PR will hopefully make our flakiest test more reliable by using randomly generated strings for entity creation (table, view and schema) in our test database.

Will hopefully help us close issue #281.

### DIFF
--- a/tests/test_mssqlcliclient.py
+++ b/tests/test_mssqlcliclient.py
@@ -121,8 +121,19 @@ class MssqlCliClientTests(unittest.TestCase):
         viewtest = "test_%s" % random_str()
         schematest = "test_%s" % random_str()
 
+        def drop_entities(mssqlcli_client):
+            list(mssqlcli_client.execute_query('DROP TABLE %s;' % tabletest1))
+            list(mssqlcli_client.execute_query('DROP TABLE %s;' % tabletest2))
+            list(mssqlcli_client.execute_query('DROP VIEW %s IF EXISTS;' % viewtest))
+            list(mssqlcli_client.execute_query('DROP TABLE %s;' % "."
+                                               .join([schematest, tabletest1])))
+            list(mssqlcli_client.execute_query('DROP SCHEMA %s;' % schematest))
+
         try:
             client = create_mssql_cli_client()
+
+            drop_entities(client)   # drop entities in beginning (in case tables exist)
+
             list(client.execute_query('CREATE TABLE %s (a int, b varchar(25));' % tabletest1))
             list(client.execute_query('CREATE TABLE %s (x int, y varchar(25), z bit);'
                                       % tabletest2))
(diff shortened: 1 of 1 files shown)


tests/test_mssqlcliclient.py
```diff
@@ -107,32 +108,42 @@ def test_json_writer_extra_params(self):
             json_writer.close()
             shutdown(client)
 
-    def test_schema_table_views_and_columns_query(self):
+    @staticmethod
+    def test_schema_table_views_and_columns_query():
```
kevcunnane: We've had this same issue elsewhere in tools. The standard solutions are:
- Spin up a new DB for each test run & dispose after, and/or:
- Run the drop statements both before and after the creates. This means that if they were left over from previous run you can remove. Otherwise you end up with a lot of crufty tables lying around when someone hits Stop halfway through a test.

Have you considered these options?

ellbosch: Thanks @kevcunnane. We've considered the first option but ran into more problems with creating more databases, including failures from lengthy test times. But I like the second option a lot, I'll add that and push out another commit.

tests/test_mssqlcliclient.py
```diff
@@ -121,8 +121,19 @@ def test_schema_table_views_and_columns_query():
...
+            list(mssqlcli_client.execute_query('DROP TABLE %s;' % tabletest2))
+            list(mssqlcli_client.execute_query('DROP VIEW %s IF EXISTS;' % viewtest))
+            list(mssqlcli_client.execute_query('DROP TABLE %s;' % "."
+                                               .join([schematest, tabletest1])))
+            list(mssqlcli_client.execute_query('DROP SCHEMA %s;' % schematest))
+
         try:
             client = create_mssql_cli_client()
+
+            drop_entities(client)   # drop entities in beginning (in case tables exist)
```
corivera: Would this cause any failures if the table doesn't exist?

ellbosch: Yes--it should probably use a try/except block.

corivera: Could we add an IF EXISTS wrapper to the utility queries instead?

corivera: Otherwise we might fail dropping the first table while others persist, which causes the test to still fail overall due to preexisting objects or something.

ellbosch: OH sorry I misunderstood your question. No, it won't fail if the table doesn't exist. But we should probably have something surrounding the client parameter in case that's set to None/not running properly.

ellbosch: On second thought, I think current functionality is ok. We don't want that function running if the client doesn't get set.

PR #329, opened 2019-11-03
PEP 8 fixes for mssql_cli.py + others

ellbosch:
This PR fixes code violating PEP 8 for mssql_cli.py and it's dependencies.

### DIFF
--- a/mssqlcli/mssql_cli.py
+++ b/mssqlcli/mssql_cli.py
@@ -76,19 +76,16 @@ def security_words_found_in(query):
     try:
         tokens = query.lower()
         return any([keyword for keyword in security_keywords if keyword in tokens])
-    except Exception:
+    except AttributeError:
         return False
 
 
 class MssqlFileHistory(FileHistory):
-    def __init__(self, filename):
-        super(self.__class__, self).__init__(filename)
-
     def append_string(self, string):
         if security_words_found_in(string):
             return
 
-        super(self.__class__, self).append_string(string)
+        super(MssqlFileHistory, self).append_string(string)
 
 
 class MssqlCli(object):
--- a/mssqlcli/mssql_cli.py
+++ b/mssqlcli/mssql_cli.py
@@ -100,11 +100,11 @@ class MssqlCli(object):
 
         if configured_pager:
             self.logger.info(
-                'Default pager found in config file: "{}"'.format(configured_pager))
+                'Default pager found in config file: "%s"', configured_pager)
             os.environ['PAGER'] = configured_pager
         elif os_environ_pager:
-            self.logger.info('Default pager found in PAGER environment variable: "{}"'.format(
-                os_environ_pager))
+            self.logger.info('Default pager found in PAGER environment variable: "%s"',
+                             os_environ_pager)
             os.environ['PAGER'] = os_environ_pager
         else:
             self.logger.info(
--- a/mssqlcli/mssql_cli.py
+++ b/mssqlcli/mssql_cli.py
@@ -451,6 +451,7 @@ class MssqlCli(object):
             prompt = self.get_prompt(self.prompt_format)
             return [(u'class:prompt', prompt)]
 
+        # FIXME: unused arguments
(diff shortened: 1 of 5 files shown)


mssqlcli/mssql_cli.py
```diff
@@ -179,14 +176,15 @@ def __init__(self, options):
             self.query_history = []
 
             # Initialize completer
-            smart_completion = True if c['main'].get('smart_completion', 'True') == 'True' else False
-            self.completer = MssqlCompleter(smart_completion=smart_completion, settings=self.settings)
+            smart_completion = c['main'].get('smart_completion', 'True') == 'True'
```
kevcunnane: Minor: is there any case where 'true' would be coming back here? Case-sensitive comparisons are always liable to have issues

mssqlcli/mssql_cli.py
```diff
@@ -269,19 +266,12 @@ def set_main_mssqlcli_client(self, mssqlcli_client):
...
-
-        try:
-            owner_uri, error_messages = self.mssqlcliclient_main.connect_to_database()
-            if not owner_uri and error_messages:
-                click.secho('\n'.join(error_messages),
-                            err=True,
-                            fg='yellow')
-                sys.exit(1)
-
-        except Exception as e:
```
kevcunnane: Where is this exception handling logic gone? What happens if an unexpected exception occurs now?

mssqlcli/mssql_cli.py
```diff
@@ -299,6 +289,7 @@ def handle_editor_command(self, text):
         # FIXME: using application.pre_run_callables like this here is not the best solution.
         # It's internal api of prompt_toolkit that may change. This was added to fix #668.
         # We may find a better way to do it in the future.
+        # pylint: disable=no-member
```
kevcunnane: So this disabling is related to the FIXME comment? If we solve, the linting issue would likely go away?

mssqlcli/mssql_cli.py
```diff
@@ -340,14 +331,11 @@ def _execute_interactive_command(self, text):
 
         except NotImplementedError:
             click.secho('Not Yet Implemented.', fg="yellow")
-        except Exception as e:
```
kevcunnane: Again, where has generic error handling moved to? Is it still cause somewhere in the stack?

mssqlcli/mssql_cli.py
ellbosch: Great point, it's unlikely but I don't think it hurts to throw in a `'.lower()`.

mssqlcli/mssql_cli.py
ellbosch: Exception handling is gone because it is already handled from upstream calls. If an exception occurs, it provides a value to the `error_messages` variable in this function.

PEP 8 also discourages using generic exception handling like what existed here. @kevcunnane what are your thoughts?

mssqlcli/mssql_cli.py
ellbosch: Yes. I created an issue to refactor the `special` package, it looks like it needs a bit of work so I scoped that out of this PR.

mssqlcli/mssql_cli.py
ellbosch: This was removed due to PEP 8 guidelines. I believe that the function has all exceptions accounted for already, but I am unsure here.

I'm not exactly tied down to doing everything PEP 8 recommends—I think some flexibility is ok, maybe we just add this back. What are your thoughts @kevcunnane?

kevcunnane:
@ellbosch do we know why integration tests failed?

ellbosch:
> @ellbosch do we know why integration tests failed?

The Windows build failure will be fixed with the upcoming PR. The Linux failure looks like one of our pre-existing flaky tests but more investigation is needed.

kevcunnane:
Given the discussion on Exception handling I'm happy to sign off since worst case is a slightly less helpful error message on exit due to unexpected errors. Please resolve integration tests but otherwise looks good

PR #331, opened 2019-11-04
fix for breaking windows test

ellbosch:
Fixes an issue where Windows insider builds have been failing due to improper use of Python subprocess calls which may exceed the buffer limit (see warning [here](https://docs.python.org/2/library/subprocess.html#subprocess.Popen.wait)).

I've also added timeout checks in case we run into tests which hang—an issue we saw with the aforementioned failing test.


ellbosch:
PR validation builds still failing, will re-explore.

Issue #283, opened 2019-09-10
Fix pep8 issues


pensivebrian:

ellbosch:
Notes:
* Automated build needs to run pylint (errors, security issues...)
* Find unused methods

ellbosch:
Related to #289.

PR #333, opened 2019-11-05
PEP 8 changes for main.py + root files

ellbosch:
Tag: #283

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,9 +2,8 @@ import ast
 import os
 import re
 import datetime
-
+from codecs import open     #pylint: disable=redefined-builtin
 from setuptools import setup, find_packages
-from codecs import open
 
 _version_re = re.compile(r'__version__\s+=\s+(.*)')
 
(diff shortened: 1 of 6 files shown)


setup.py
```diff
@@ -2,9 +2,8 @@
 import os
 import re
 import datetime
-
+from codecs import open     #pylint: disable=redefined-builtin
```
kevcunnane: Is there a reason we don't just import this as a different name and then do that? E.g. call it codec_open or something?

ellbosch: Upon revisiting, we might be able to remove this altogether. I'll test this first and send out a commit if it works.

ellbosch: After testing this with the native `open` module, we may want to leave this as is. By importing `codecs`, we maintain consistency between Python 2 and 3. This is because the `open` module doesn't support encoding type in Python 2, however, `codecs` does. We need to set encoding type because Mac and Windows use different ones.

@kevcunnane I like this method because it overrides the native `open` module, which ensures we'll use the `codecs` version. What are your thoughts?

PR #334, opened 2019-11-05
migrated tests off of deprecated isAlive methods

ellbosch:
We've had builds fail from calls made with `.response.thread.isAlive()` which are deprecated methods. Hopefully this will help mitigate future failures.

Tag: #281


ellbosch:
> This doesn't look likely to actively hurt anything, but do we think it'll make a positive change? Do we need to do some kind of wait / yield the main process so the thread can shut itself down instead?

I'm hoping it will, tests are passing on this branch so it doesn't seem to be introducing breaking changes. If we see more failures post-merge then we'll know there's more work to do.

It's super hard to repro because it almost always passes, it's the occasional failure that gets us.

PR #335, opened 2019-11-07
improved debugging and pep 8 support for globalization tests

ellbosch:
Our Windows build recently failed during one of our globalization tests. This is a known flaky test--however, there isn't any useful stack trace to help us debug the failures due to generic exception handling.

Therefore, I removed generic exception handling so that errors are no longer suppressed. Furthermore, I used this opportunity to convert unit tests to Pytest (a more modern version of Python's built-in unit test module) as well as add PEP 8 compatibility.

Tags: #281, #283

### DIFF
--- a/tests/test_globalization.py
+++ b/tests/test_globalization.py
@@ -170,7 +170,7 @@ class TestGlobalizationMetadata(GlobalizationTests):
     def test_db():
         db = create_test_db()
         yield db
-        return clean_up_test_db(db)
+        clean_up_test_db(db)
 
     def test_schema_metadata_double(self, test_db):
         self.run_schema_metadata_validation('double', test_db)
--- a/tests/test_globalization.py
+++ b/tests/test_globalization.py
@@ -1,6 +1,6 @@
 # coding=utf-8
-import unittest
 import socket
+import pytest
 from prompt_toolkit.document import Document
 from utility import random_str
 from mssqltestutils import (
@@ -14,25 +14,19 @@ from mssqlcli.mssqlcompleter import MssqlCompleter
 import mssqlcli.completion_refresher as completion_refresher
 
 
-"""
-All tests require a live connection to test server.
-Make modifications to mssqlutils.create_mssql_cli_client()
-to use a different server and database.
-Please Note: These tests cannot be run offline.
-"""
+# All tests require a live connection to test server.
+# Make modifications to mssqlutils.create_mssql_cli_client()
+# to use a different server and database.
+# Please Note: These tests cannot be run offline.
 
 
-class GlobalizationTests():
+class GlobalizationTests:
     @staticmethod
     def run_query(mssqlcliclient, query):
         success = True
-        try:
-            for _, _, _, _, is_error in mssqlcliclient.execute_query(query):
-                if is_error is True:
-                    success = False
-                    break
-        except:
-            success = False
+        for _, _, status, _, is_error in mssqlcliclient.execute_query(query):
+            if is_error is True:
+                raise AssertionError(status)
         return success
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


ellbosch:
Looks like I was handling cleanup incorrectly with Python 2 versions--will commit a fix here.

ellbosch:
Got the error again, this time in Linux. Last time it was Windows, so there doesn't appear to be an OS-related issue here.

The `special` package occasionally throws an error saying the command isn't found for a `CREATE SCHEMA` query. I've only seen this happen with the globalizations tests, so I'm guessing it may have something to do with that. Furthermore, the only test I've seen fail is the first one in the tests for result set.

I'm unsure why this happens only some of the time. With my last commit, we'll have more info on the parsed command that isn't getting found.

tests/test_globalization.py
```diff
@@ -14,25 +14,19 @@
...
     def run_query(mssqlcliclient, query):
         success = True
-        try:
-            for _, _, _, _, is_error in mssqlcliclient.execute_query(query):
-                if is_error is True:
-                    success = False
-                    break
-        except:
-            success = False
+        for _, _, status, _, is_error in mssqlcliclient.execute_query(query):
```
kevcunnane: OK so this captures additional details from the call now and logs in the exception stack?

tests/test_globalization.py
```diff
@@ -77,35 +71,37 @@ def get_next_characters(all_characters, count=50):
             u'㔤㔥㔦㔧㔨㔩㔪㔫㔬㔭㔮㔯㔰㔱㔲㔳㔴㔵㔶㔷㔸㔹㔺㔻㔼㔽㔾㔿㕀㕁㕂㕃㕄㕅㕆㕇㕈㕉㕊㕋㕌㕍㕎㕏㕐㕑㕒㕓㕔㕕㕖㕗㕘㕙㕚㕛㕜㕝'
         ),
         'mongolian': (
-            u'᠐᠑᠒᠓᠔᠕᠖᠗᠘᠙ᠠᠡᠢᠣᠤᠥᠦᠧᠨᠩᠪᠫᠬᠭᠮᠯᠰᠱᠲᠳᠴᠵᠶᠷᠸᠹᠺᠻᠼᠽᠾᠿᡀᡁᡂᡃᡄᡅᡆᡇᡈᡉᡊᡋᡌᡍᡎᡏᡐᡑᡒᡓᡔᡕᡖᡗᡘᡙᡚᡛᡜᡝᡞᡟᡠᡡᡢᡣᡤᡥᡦᡧᡨᡩᡪᡫᡬᡭᡮᡯ'
-            u'ᡰᡱᡲᡳᡴᡵᡶᡷᢀᢁᢂᢃᢄᢅᢆᢇᢈᢉᢊᢋᢌᢍᢎᢏᢐᢑᢒᢓᢔᢕᢖᢗᢘᢙᢚᢛᢜᢝᢞᢟᢠᢡᢢᢣᢤᢥᢦᢧᢨᢩ'
+            u'᠐᠑᠒᠓᠔᠕᠖᠗᠘᠙ᠠᠡᠢᠣᠤᠥᠦᠧᠨᠩᠪᠫᠬᠭᠮᠯᠰᠱᠲᠳᠴᠵᠶᠷᠸᠹᠺᠻᠼᠽᠾᠿᡀᡁᡂᡃᡄᡅᡆᡇᡈᡉᡊᡋᡌᡍᡎᡏᡐᡑᡒᡓᡔᡕᡖᡗᡘᡙᡚᡛᡜᡝᡞᡟᡠᡡᡢᡣᡤᡥᡦᡧᡨᡩᡪ'
```
kevcunnane: Why do these strings show as different? Is there some change that happened when pulling these down?

tests/test_globalization.py
```diff
@@ -168,61 +164,60 @@ def run_charset_validation(self, charset):
...
 
-class GlobalizationMetadataTests(GlobalizationTests, unittest.TestCase):
-    @classmethod
-    def setUpClass(cls):
-        cls.test_db = create_test_db()
-
-    @classmethod
-    def tearDownClass(cls):
-        clean_up_test_db(cls.test_db)
+class TestGlobalizationMetadata(GlobalizationTests):
```
kevcunnane: Remind me: Does pytest detect tests by convention? E.g. it finds anything starting with Test and treats as test class?

tests/test_globalization.py
```diff
@@ -168,61 +164,60 @@ def run_charset_validation(self, charset):
...
+    @staticmethod
+    @pytest.fixture(scope="class")
+    def test_db():
+        db = create_test_db()
+        yield db
```
kevcunnane: Is this like a try/finally implementation? If error is thrown / test fails, does the cleanup still happen?

tests/test_globalization.py
kevcunnane: This and the change to main.py are the only 2 required to identify why the test fails, correct?

ellbosch: Yes, all correct.

tests/test_globalization.py
ellbosch: The strings are the same, with the exception of some newlines I added to suppress warnings about line widths.

tests/test_globalization.py
ellbosch: Exactly as you described--`test` must prefix class/function names.

tests/test_globalization.py
ellbosch: Pytest fixtures appear to be an improved version over the classic setup and teardown style methods of unit tests. Fixtures ensure cleanup as well. More info [here](https://docs.pytest.org/en/latest/fixture.html) if you're curious.

ellbosch:
> This looks OK in terms of implementation. 2 things for the future:
> 
> * Please separate out change to test framework from the specific code changes to identify problem. Combining makes it hard to track
> * I'm wondering whether changes like this make us move away from what the other dbcli tools use for code, framework etc. Are we trying to align with them in any way? Can we build on them in terms of test reliability, what tests to run, etc.?
> 
> I'd prefer if we split this into the 2 PRs just so they're clean, but OK with this as-is.

Fair point on splitting this up into different PRs--I definitely got a little carried away here. I'll make sure to do that in the future.

Alignment with other dbcli's is something we can do better. I can reach out to the folks to see what they do.

PR #336, opened 2019-11-09
PEP 8 changes in mssqlcli folder

ellbosch:
Tag: #283

### DIFF
--- a/mssqlcli/completion_refresher.py
+++ b/mssqlcli/completion_refresher.py
@@ -62,8 +62,8 @@ class CompletionRefresher:
             callbacks = [callbacks]
 
         while 1:
-            for r in self.refreshers.values():
-                r(completer, executor)
+            for refresh in self.refreshers.values():
+                refresh(completer, executor)
                 if self._restart_refresh.is_set():
                     self._restart_refresh.clear()
                     break
--- a/mssqlcli/completion_refresher.py
+++ b/mssqlcli/completion_refresher.py
@@ -1,14 +1,13 @@
+from collections import OrderedDict
 import logging
 import threading
 import mssqlcli.decorators as decorators
-
-from collections import OrderedDict
 from .mssqlcompleter import MssqlCompleter
 
 logger = logging.getLogger(u'mssqlcli.completion_refresher')
 
 
-class CompletionRefresher(object):
+class CompletionRefresher:
 
     refreshers = OrderedDict()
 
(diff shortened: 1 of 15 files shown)


chlafreniere:
Can you help me understand which changes are required for PEP 8 compliance and which aren't? 

For example, removing a parameter from a constructor (line 10 in the first file) doesn't seem to fix anything on LGTM.

ellbosch:
> Can you help me understand which changes are required for PEP 8 compliance and which aren't?
> 
> For example, removing a parameter from a constructor (line 10 in the first file) doesn't seem to fix anything on LGTM.

Truthfully, it's PEP 8 plus other stuff. General code cleanup. I'm going through each issue that's picked up from pylint, which includes PEP 8.

mssqlcli/mssqlcompleter.py
```diff
@@ -429,6 +433,8 @@ def case(self, word):
         return self.casing.get(word, word)
 
     def get_completions(self, document, complete_event, smart_completion=None):
+        # pylint: disable=arguments-differ
```
kevcunnane: nitpick: do we need the newline after this?

utility.py
```diff
@@ -83,24 +83,5 @@ def copy_current_platform_mssqltoolsservice():
         print("This platform: {} does not support mssqltoolsservice.".format(platform.system()))
 
 
-def encode(s):
```
kevcunnane: These were unused?

mssqlcli/completion_refresher.py
```diff
@@ -63,8 +62,8 @@ def _bg_refresh(self, mssqlcliclient, callbacks, history=None,
             callbacks = [callbacks]
 
         while 1:
-            for refresher in self.refreshers.values():
-                refresher(completer, executor)
+            for r in self.refreshers.values():
```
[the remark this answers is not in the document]
ellbosch: This was changed was because `refresher` is also a package name that's being used here.

utility.py
ellbosch: Correct. We moved these methods to `mssqlcli/util.py` a few weeks back.

mssqlcli/completion_refresher.py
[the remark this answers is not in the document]
ellbosch: Just renamed to `refresh` so hopefully that's a little better now.

mssqlcli/mssqlcompleter.py
ellbosch: Just my opinion, I don't think it hurts. I think it helps group the comment with the function definition.

PR #351, opened 2019-11-19
PEP 8 changes

ellbosch:
Tag: #283.

Looks like there's quite a few files I missed for PEP 8 changes. Rather than submit 1 giant PR, I will send out smaller ones.

The biggest change here is to [generatequerybaseline.py](https://github.com/dbcli/mssql-cli/compare/ellbosch/pep8-2?expand=1#diff-4a78448e7b7a448863fae4fc4d412999). The `id` keyword was renamed to `request_id` since Python already has a built-in `id` variable. This impacted a few files leveraging these methods, since the keywords (if used) needed to be renamed.


kevcunnane:
Is the failed test flaky? Any idea the cause?

mssqlcli/mssqlcliclient.py
```diff
@@ -20,7 +19,8 @@ def generate_owner_uri():
     return u'mssql-cli-' + uuid.uuid4().urn
 
 
-class MssqlCliClient(object):
+class MssqlCliClient:
+    # pylint: disable=too-many-instance-attributes
```
kevcunnane: Just a consideration: what pylint rules are we using? Are they mandated by a standard Microsoft process? I'm wondering whether for some classes of issue, relaxing the rules is better than fixing / spot-ignoring them in each place.

ellbosch: We're using pylint's default configuration. I don't know of anything that's Microsoft mandated.

It is possible to relax the rules, however, I generally like the pylint suggestions moving forward. Even with this specific commented line, where I didn't feel like refactoring the class for the sake of this issue, I think that moving forward if we adhere to these rules, the better our code will be.

With that being said, it's definitely worth having a discussion with Brian (when he's back) about what rules we'd like to ignore moving forward.

ellbosch:
> Is the failed test flaky? Any idea the cause?

Flaky test. I need to investigate this further.

ellbosch:
The build failure is the result of a connection timeout on a query. This is the first time this test has failed on a network error, and has been stable for some time. I'll watch this closely.

PR #355, opened 2019-11-21
PEP 8 changes in JSON RPC

ellbosch:
Tag: #351 

More PEP 8 changes, this time focusing on JSON RPC module. Roughly 65 pylint warnings are fixed in this PR, with an estimated 180 which remain.

### DIFF
--- a/mssqlcli/jsonrpc/contracts/__init__.py
+++ b/mssqlcli/jsonrpc/contracts/__init__.py
@@ -16,15 +16,18 @@ class Request(ABC):
         """
             Executes the request.
         """
+        pass
 
     @abc.abstractmethod
     def get_response(self):
         """
             Retrieves expected response.
         """
+        pass
 
     @abc.abstractmethod
     def completed(self):
         """
             Return state of request.
         """
+        pass
--- a/mssqlcli/jsonrpc/contracts/__init__.py
+++ b/mssqlcli/jsonrpc/contracts/__init__.py
@@ -16,18 +16,15 @@ class Request(ABC):
         """
             Executes the request.
         """
-        pass
 
     @abc.abstractmethod
     def get_response(self):
         """
             Retrieves expected response.
         """
-        pass
 
     @abc.abstractmethod
     def completed(self):
         """
             Return state of request.
         """
-        pass
--- a/mssqlcli/jsonrpc/contracts/queryexecutestringservice.py
+++ b/mssqlcli/jsonrpc/contracts/queryexecutestringservice.py
@@ -1,3 +1,6 @@
+# pylint: disable=too-many-instance-attributes
+# pylint: disable=too-few-public-methods
+
 import logging
 from mssqlcli.jsonrpc.contracts import Request
 
@@ -11,8 +14,8 @@ class QueryExecuteStringRequest(Request):
 
     METHOD_NAME = u'query/executeString'
 
-    def __init__(self, id, owner_uri, json_rpc_client, parameters):
-        self.id = id
+    def __init__(self, request_id, owner_uri, json_rpc_client, parameters):
+        self.request_id = request_id
         self.owner_uri = owner_uri
         self.finished = False
         self.json_rpc_client = json_rpc_client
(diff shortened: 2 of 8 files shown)


mssqlcli/jsonrpc/contracts/tests/test_json_rpc_contracts.py
```diff
@@ -214,7 +219,8 @@ def verify_query_service_response(self,
                         query_batch_summaries = len(response.batch_summaries)
                         query_result_set_summaries = len(
                             response.batch_summaries[0].result_set_summaries)
-                        query_row_count = response.batch_summaries[0].result_set_summaries[0].row_count
+                        query_row_count = response.batch_summaries[0].result_set_summaries[0]\
```
corivera: Why this one? Is there a max line length rule?

mssqlcli/jsonrpc/contracts/__init__.py
```diff
@@ -16,18 +16,15 @@ def execute(self):
...
             Executes the request.
         """
-        pass
 
     @abc.abstractmethod
     def get_response(self):
         """
             Retrieves expected response.
         """
-        pass
```
kevcunnane: Looking online it seems like using pass is the standard way of defining empty methods. Why are we having to remove it? What's the rule it fixes?

mssqlcli/jsonrpc/contracts/queryexecutestringservice.py
```diff
@@ -23,50 +26,51 @@ def get_response(self):
             Get latest response, event or exception if occured.
         """
         try:
-            response = self.json_rpc_client.get_response(self.id, self.owner_uri)
+            response = self.json_rpc_client.get_response(self.request_id, self.owner_uri)
 
             decoded_response = None
             if response:
                 decoded_response = self.decode_response(response)
```
kevcunnane: Given this is now static, should't we use the class and not `self` to call it?

mssqlcli/jsonrpc/contracts/queryexecutestringservice.py
```diff
@@ -153,37 +157,42 @@ def completed(self):
 
     def get_response(self):
         try:
-            response = self.json_rpc_client.get_response(self.id, self.owner_uri)
+            response = self.json_rpc_client.get_response(self.request_id, self.owner_uri)
             decoded_response = None
             if response:
                 decoded_response = self.decode_response(response)
```
kevcunnane: Again isn't this now static and should be called on the class not self?

mssqlcli/jsonrpc/tests/test_jsonrpcclient.py
```diff
@@ -118,10 +120,15 @@ def test_send_multiple_request(self):
         self.shutdown_background_threads(test_client)
```
kevcunnane: Another case where we should call this static method on the class not `self`?

ellbosch:
Sorry everyone--I should have completed more tests before adding folks to review. I'll re-assess and send an update here.

mssqlcli/jsonrpc/contracts/tests/test_json_rpc_contracts.py
ellbosch: Yep

mssqlcli/jsonrpc/contracts/__init__.py
ellbosch: `pass` statements are no longer necessary but in this case I think I'd be ok re-including it and adding it as a rule to ignore. Thoughts @kevcunnane?

mssqlcli/jsonrpc/contracts/queryexecutestringservice.py
ellbosch: Interesting question... I suppose we could. Would there be any effective difference? Genuinely curious here.

ellbosch: After some thought, I like your idea of referencing the class instead of `self` for static methods. For readability's sake, it makes it clear and obvious that it's a static method being used.

ellbosch:
The test failure is due to a flaky test. Otherwise, this PR is ready for review.

PR #360, opened 2019-11-22
PEP 8 changes

ellbosch:
Last annoying PR for PEP 8! Closes #351.

You may notice that mssqlcli/packages/sqlcompletion.py is a bit of a mess--I've handled most PEP 8 issues here but I raised a separate issue to properly refactor it.


ellbosch:
Will explore test failures and send update.

PR #369, opened 2019-11-27
Added retry logic for create db

### LINKED ISSUE #280: Create database against Azure needs to more resilient.
If the cloud is having a bad day, a create database statement may fail with a command timeout, but the operation may be in-progress. We should check the server sys.dm_operation_status DMV for the status of the create database operation, and poll until the create database operation completes.

ellbosch:
Fixes #280.

I added retry logic to the create db method used for our tests. In the past this method would occasionally fail due to network issues, so I'm hoping that the retry logic will fix this issue.

### DIFF
--- a/tests/mssqltestutils.py
+++ b/tests/mssqltestutils.py
@@ -78,12 +78,30 @@ def create_test_db():
         '-', '_').replace('.', '_')
     test_db_name = u'mssqlcli_testdb_{0}_{1}'.format(
         local_machine_name, random_str())
-    query = u"CREATE DATABASE {0};".format(test_db_name)
-    for _, _, _, _, is_error in client.execute_query(query):
-        if is_error is True:
-            test_db_name = None
+    query_db_create = u"CREATE DATABASE {0};".format(test_db_name)
+    rows, columns, status, statement, is_error = None, None, None, None, None
+    count = 0
+
+    # retry logic in case db create fails
+    while count < 5:
+        for _, _, _, _, is_create_error in client.execute_query(query_db_create):
+            if not is_create_error:
+                shutdown(client)
+                return test_db_name
+        count += 1
     shutdown(client)
-    return test_db_name
+
+    # cleanup db just in case, then raise exception
+    clean_up_test_db(test_db_name)
+    raise AssertionError(
+        ("DB creation failed:\n"
+         "rows:\t{0}\n"
+         "columns:\t{1}\n"
+         "status:\t{2}\n"
+         "statement:\t{3}\n"
+         "is_error:\t{4}").format(rows, columns, status, statement, is_error)
+    )
+
 
 def clean_up_test_db(test_db_name):
     client = create_mssql_cli_client()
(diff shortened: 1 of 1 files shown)


tests/mssqltestutils.py
```diff
@@ -78,12 +78,30 @@ def create_test_db():
...
-        if is_error is True:
-            test_db_name = None
+    query_db_create = u"CREATE DATABASE {0};".format(test_db_name)
+    rows, columns, status, statement, is_error = None, None, None, None, None
+    count = 0
+
+    # retry logic in case db create fails
+    while count < 5:
+        for _, _, _, _, is_create_error in client.execute_query(query_db_create):
+            if not is_create_error:
```
chlafreniere: what if it is a create error? can you try cleaning up? i'm curious what happens if a database just takes longer to create than whatever timeout is set.

in addition, when cleaning up, you probably need to cleanup all test database names, instead of just the last one, if i'm understanding this code correctly.

udeeshagautam: that is a good point. 

@ellbosch one more small thing - should we try to check for online as well as we discussed to be sure it is "select count(1) from sys.databases where state = 0 and name = '...'

ellbosch: > @ellbosch one more small thing - should we try to check for online as well as we discussed to be sure it is "select count(1) from sys.databases where state = 0 and name = '...'

I was unsuccessful at checking state because the database wasn't appearing in sys.databases. It's a weird issue. It might be related to the fact that we're not logged in to the master database when the create db function is called. I'm really unsure, but I hope that my approach will still be sufficient.

ellbosch: > i'm curious what happens if a database just takes longer to create than whatever timeout is set.

This is a very good point, which I had to go look up and see if this is possible. From what I can deduce, there is no such query that will return before its execution is complete (as seen [here](https://github.com/dbcli/mssql-cli/blob/master/mssqlcli/mssqlcliclient.py#L223)). Therefore, I will need to update the PR so that query execution is reattempted.

> in addition, when cleaning up, you probably need to cleanup all test database names, instead of just the last one, if i'm understanding this code correctly.

Because this code is only in charge of creating a singular testing db, we just want to delete that one only.

ellbosch: Correction on my previous comment. The query execution actually does reattempt on each failure. After 5 failures, the method raises an assertion error and a cleanup method gets called on the DB. Given this I believe this PR is good as is, but please feel free to provide more feedback should I be missing something

chlafreniere: oh duh ok it's the same database name every time. my bad!

chlafreniere: how about this... if it is a create error, can we log the error message, so we know what's gone wrong?

ellbosch: Like this. I'll update.

ellbosch:
Latest commit will log a warning whenever a DB create execution fails.

I also modified the error message at the end to no longer print values since they were all set to `None` anyways, but thankfully pytest already prints all local variables to stderror.

Issue #371, opened 2019-11-28
JSON RPC client test function needs to be fixed


ellbosch:
The `test_stream_closed_during_process` method does not throw an exception when it should. This issue was not caught because the test method wouldn't raise a failure if the exception block wasn't run (which is expected functionality).

Further investigation is needed to understand why the exception isn't raised.

PR #372, opened 2019-11-28
JSON RPC client testing improvements

ellbosch:
Happy Thanksgiving!

This PR improves our testing with JSON RPC client. Changes specifically address feedback from the Weekly Tech Review.

Changes:
* Addresses a hopeful fix for the flaky test that sometimes failed to shut down a thread.
* Raises assertion errors when exceptions should've been caught but didn't.

In thanks to these changes, I discovered that one of our tests fails to catch an exception when it should've. I raised an issue at #371 and have marked that specific test method to skip on runtime.

Tag: #281

### DIFF
--- a/mssqlcli/jsonrpc/tests/test_jsonrpcclient.py
+++ b/mssqlcli/jsonrpc/tests/test_jsonrpcclient.py
@@ -1,6 +1,7 @@
 import unittest
 import time
 import io
+import threading
 
 import mssqlcli.jsonrpc.jsonrpcclient as json_rpc_client
 
@@ -188,11 +189,27 @@ class JsonRpcClientTests(unittest.TestCase):
             self.assertEqual(
                 str(exception),
                 u'Content-Length was not found in headers received.')
+
+            # flaky test fix: logic that waits until threads shut down
+            count = 0
+            while count < 5:
+                active_threads = threading.enumerate()
+                if test_client.request_thread not in active_threads or \
+                    test_client.response_thread in active_threads:
+                    time.sleep(5)
+                    count += 1
+                else:
+                    break
+
             # Lookup exception for invalid content length spelling.
             self.assertTrue(test_client.request_thread.is_alive())
             self.assertFalse(test_client.response_thread.is_alive())
             test_client.shutdown()
             self.assertFalse(test_client.request_thread.is_alive())
+        else:
+            raise AssertionError("LookupError exception not caught when it should've been.")
+        finally:
+            test_client.shutdown()
 
     def test_response_stream_closed_exception(self):
         """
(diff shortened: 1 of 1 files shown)


mssqlcli/jsonrpc/tests/test_jsonrpcclient.py
```diff
@@ -188,11 +189,27 @@ def test_receive_invalid_response_exception(self):
...
                 u'Content-Length was not found in headers received.')
+
+            # flaky test fix: logic that waits until threads shut down
+            count = 0
+            while count < 5:
+                active_threads = threading.enumerate()
+                if test_client.request_thread not in active_threads or \
+                    test_client.response_thread in active_threads:
+                    time.sleep(5)
+                    count += 1
```
udeeshagautam: what if we pass 5 and thread is still not shutdown?

mssqlcli/jsonrpc/tests/test_jsonrpcclient.py
```diff
@@ -260,6 +280,9 @@ def test_stream_closed_during_process(self):
                 str(exception), u'I/O operation on closed file.')
             # Verify response thread is dead.
             self.assertFalse(test_client.request_thread.is_alive())
```
udeeshagautam: just for my understanding... in first test (line 207-208) we shutdown and then check for thread.. and here we check before shutdown. Why the difference?

udeeshagautam:
You might want to re-run PR build

mssqlcli/jsonrpc/tests/test_jsonrpcclient.py
ellbosch: Not positive. In any case it's a broken test, should sync with Brian about this when he's back.

mssqlcli/jsonrpc/tests/test_jsonrpcclient.py
ellbosch: Then the test will likely fail, because the following assertion checks for a request thread that has likely not been created.

The only way the request thread _could_ have been created following 5 failed retries is if it happens to come alive following the fifth try. I have not seen this happen during my testing.

mssqlcli/jsonrpc/tests/test_jsonrpcclient.py
```diff
@@ -188,11 +189,27 @@ def test_receive_invalid_response_exception(self):
...
                 str(exception),
                 u'Content-Length was not found in headers received.')
+
+            # flaky test fix: logic that waits until threads shut down
+            count = 0
+            while count < 5:
+                active_threads = threading.enumerate()
+                if test_client.request_thread not in active_threads or \
+                    test_client.response_thread in active_threads:
+                    time.sleep(5)
```
chlafreniere: This is 5 seconds, right? Are there any timeouts on the tests themselves?

ellbosch: Yes, 5 seconds. No timeouts from what I can tell.

PR #382, opened 2019-12-05
Add support for unstable unit tests

### LINKED ISSUE #381: Move flaky tests to separate build task
This is to help us get to all green builds passing while flaky tests continue to get fixed. Should help us close #281.

ellbosch:
Closes #381.

To help builds pass more frequently, we need to separate flaky tests from our normal unit testing. To do this, I've created a new build function to separately run marked `unstable` tests. I've also added new tasks to our PR builds to test unstable unit tests from our Azure pipelines.

### DIFF
--- a/build.py
+++ b/build.py
@@ -126,10 +126,23 @@ def unit_test():
     """
     runid = str(uuid.uuid1())
     python_version = platform.python_version()
-    utility.exec_command(
-        'pytest -l --cov mssqlcli --doctest-modules --junitxml=junit/test-{}-results.xml \
-            --cov-report=xml --cov-report=html --cov-append '
-        '-o junit_suite_name=pytest-{} '
+    utility.exec_command(('pytest -l --cov mssqlcli --doctest-modules '
+                          '--junitxml=junit/test-{}-results.xml --cov-report=xml '
+                          '--cov-report=html --cov-append -o junit_suite_name=pytest-{} '
+                          '-s -m "not unstable" {}').format(runid, python_version,
+                                                            get_active_test_filepaths()),
+                         utility.ROOT_DIR, continue_on_error=False)
+
+def unstable_unit_test():
+    """
+    Run all unstable unit tests.
+    """
+    utility.exec_command('pytest -s -v -m unstable {}'.format(get_active_test_filepaths()),
+                         utility.ROOT_DIR, continue_on_error=False)
+
+
+def get_active_test_filepaths():
+    return (
         'tests/test_mssqlcliclient.py '
         'tests/test_completion_refresher.py '
         'tests/test_config.py '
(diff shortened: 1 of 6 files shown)


utility.py
```diff
@@ -21,7 +22,7 @@ def exec_command(command, directory, continue_on_error=True):
         Execute command.
     """
     try:
-        check_call(command.split(), cwd=directory)
+        check_call(shlex.split(command, posix=False), cwd=directory)
```
ellbosch: By leveraging `shlex` (a native Python module) we're able to process commands with quotations as well as retain the quotations using `posix=False`. This is necessary for our new call to stable unit tests that needs to add `"not unstable"` to the testing command line call.

build.py
```diff
@@ -126,10 +126,23 @@ def unit_test():
...
+                          '-s -m "not unstable" {}').format(runid, python_version,
+                                                            get_active_test_filepaths()),
+                         utility.ROOT_DIR, continue_on_error=False)
+
+def unstable_unit_test():
+    """
+    Run all unstable unit tests.
+    """
+    utility.exec_command('pytest -s -v -m unstable {}'.format(get_active_test_filepaths()),
+                         utility.ROOT_DIR, continue_on_error=False)
```
ellbosch: The new function call for unstable tests is much more concise than the stable call because it's not concerned with code coverage.

As an aside, I added a `-v` argument to make test logs more verbose since there are fewer unstable tests.

build.py
```diff
@@ -126,10 +126,23 @@ def unit_test():
     """
     runid = str(uuid.uuid1())
     python_version = platform.python_version()
-    utility.exec_command(
-        'pytest -l --cov mssqlcli --doctest-modules --junitxml=junit/test-{}-results.xml \
-            --cov-report=xml --cov-report=html --cov-append '
-        '-o junit_suite_name=pytest-{} '
+    utility.exec_command(('pytest -l --cov mssqlcli --doctest-modules '
```
chlafreniere: do you have a test run that you can show with unstable tests running separately?

build.py
```diff
@@ -126,10 +126,23 @@ def unit_test():
...
+                          '--junitxml=junit/test-{}-results.xml --cov-report=xml '
+                          '--cov-report=html --cov-append -o junit_suite_name=pytest-{} '
+                          '-s -m "not unstable" {}').format(runid, python_version,
```
chlafreniere: is it possible to run with a ! (or some other character) instead of "not" to get rid of the code below?

build.py
ellbosch: Yep, here's an [example](https://mssqltools.visualstudio.com/cli/_build/results?buildId=49027&view=results) of a test where unstable tests actually failed.

build.py
ellbosch: Gave ! a shot, no dice. [Docs](http://doc.pytest.org/en/latest/example/markers.html) don't suggest of any alternatives.

PR #384, opened 2019-12-06
Create source distributions on macOS during build

### LINKED ISSUE #322: Create an sdist distribution
Needed for brew, may be needed for azdata.

ellbosch:
Closes #322.

This PR adds support for creating gztar and zip source distributions. The `sdist` call is made during the build phase, which is also when wheel files are created.

Furthermore, I added a check to only call `sdist` if the platform is macOS. This is because unlike wheel files, it is not required to generate source distributions on every platform we wish to support*, since we can generate both the gztar and zip from macOS.

*I need to test this hypothesis :)


chlafreniere:
Please make sure the checks pass 😄

build.py
```diff
@@ -80,6 +80,11 @@ def build():
                              utility.ROOT_DIR,
                              continue_on_error=False)
 
+        # generate sdist--only run on Mac to eliminate redundant copies when published to Azure
+        if sys.platform == 'darwin':
+            utility.exec_command('%s setup.py check -r -s sdist --formats=gztar,zip' % PYTHON,
+                                 utility.ROOT_DIR, continue_on_error=False)
+
```
chlafreniere: please test the entire product E2E with these changes

chlafreniere: maybe by uploading to pypi test 😄

build.py
```diff
@@ -80,6 +80,11 @@ def build():
                              utility.ROOT_DIR,
                              continue_on_error=False)
 
+        # generate sdist--only run on Mac to eliminate redundant copies when published to Azure
+        if sys.platform == 'darwin':
```
chlafreniere: please document that mac wasn't chosen for a reason (you just needed to choose one platform) to avoid future confusion

ellbosch:
@chlafreniere: a few changes made since our meeting yesterday:
* zip file option removed (in favor of targz) since only one sdist can be supported
* minor changes to variable names to make things clearer
* version bump

release.py
```diff
@@ -100,13 +100,13 @@ def publish_official():
     """
     Publish mssql-cli package to PyPi.
     """
-    mssqlcli_wheel_dir = os.listdir(utility.MSSQLCLI_DIST_DIRECTORY)
+    mssqlcli_dist_dir = os.listdir(utility.MSSQLCLI_DIST_DIRECTORY)
     # Run twine action for mssql-cli.
     # Only authorized users with credentials will be able to upload this package.
     # Credentials will be stored in a .pypirc file.
-    for wheel in mssqlcli_wheel_dir:
+    for f in mssqlcli_dist_dir:
```
corivera: What is "f" here?

ellbosch: Stands for file. I wanted to disambiguate this from wheels since we have an sdist now.

corivera: Could we use "file" here instead, to avoid ambiguity?

ellbosch: No, unfortunately, since `file` is a builtin. I'm open to a better name here--maybe `file_dist`?

PR #396, opened 2019-12-17
Multi-line statements now execute on 'GO'

### LINKED ISSUE #175: Do not use a semicolon to end a multi-line query
Currently when editing a multiline query, we use use a semicolon+<enter> to delineate the query should be executed.

Feedback from the MVP summit: in multiline query mode, users like to enter more that one statement, which are delineated by semicolons, so using a semicolon is a bad choice. We should consider using the 'GO' keyword to execute the query, which aligns with sqlcmd.

ellbosch:
Closes #175.

Changes include:
* Multi-line queries only execute when `GO` is the only token on a newline
* Updated all code references for using `GO` instead of semicolon
* Bug fix that caused some multi-line statements to crash when batching multiple queries
* Tests for valid multi-line completion

**Disclaimer:** I have not accounted for an edge cause where a block comment may end on the same line as `GO`. For example, the following statement should execute but does not for us:
```
select 1 /* hey some comment
and the comment keeps going */ go
```

I have decided to forgo this edge case in favor of waiting for our dependency, sqlparse, to fix a bug that strips newlines from a method I'm using to strip comments (tracked [here](https://github.com/andialbrecht/sqlparse/issues/484)). I'm open to a discussion on this @udeeshagautam @chlafreniere.

### DIFF
--- a/mssqlcli/mssqlbuffer.py
+++ b/mssqlcli/mssqlbuffer.py
@@ -37,10 +37,13 @@ def _is_complete(sql):
         # check that 'go' is only token on newline
         lines = sql.split('\n')
         lastline = lines[len(lines) - 1].lower().strip()
-        is_valid_go_on_lastline = 'go' == lastline
+        is_valid_go_on_lastline = lastline == 'go'
 
         # check that 'go' is on last line, not in open quotes, and there's no open
-        # comment with closed comments and quotes removed
+        # comment with closed comments and quotes removed.
+        # NOTE: this method fails when GO follows a closing '*/' block comment on the same line,
+        # we've taken a dependency with sqlparse
+        # (https://github.com/andialbrecht/sqlparse/issues/484)
         return not is_open_quote(sql) and not is_open_comment and is_valid_go_on_lastline
 
     return False
--- a/mssqlcli/mssqlbuffer.py
+++ b/mssqlcli/mssqlbuffer.py
@@ -24,7 +24,9 @@ def _is_complete(sql):
     # A complete command is an sql statement that ends with a 'GO', unless
     # there's an open quote surrounding it, as is common when writing a
     # CREATE FUNCTION command
-    return sql.lower().endswith('go') and not is_open_quote(sql)
+    if sql is not "":
+        tokens = sql.split()
+        return tokens[len(tokens) - 1].lower() == 'go' and not is_open_quote(sql)
 
 
 def _multiline_exception(text):
--- a/mssqlcli/mssqlbuffer.py
+++ b/mssqlcli/mssqlbuffer.py
@@ -24,17 +24,19 @@ def _is_complete(sql):
     # A complete command is an sql statement that ends with a 'GO', unless
     # there's an open quote surrounding it, as is common when writing a
     # CREATE FUNCTION command
-    if sql is not "":
+    if sql != "":
         tokens = sql.split('\n')
         lastline = tokens[len(tokens) - 1]
-        
+
         # remove commented code in last line to ensure we have a valid 'go'
         # to end the multiline query
-        comments = re.findall('\/\*.*?\*\/|--.*?\n', lastline)
+        comments = re.findall(r'\/\*.*?\*\/|--.*', lastline)
         for comment in comments:
             lastline = lastline.replace(comment, '')
 
         return lastline.lower().strip() == 'go' and not is_open_quote(sql)
+
+    return False
 
 
 def _multiline_exception(text):
--- a/mssqlcli/mssqlbuffer.py
+++ b/mssqlcli/mssqlbuffer.py
@@ -21,10 +21,10 @@ def mssql_is_multiline(mssql_cli):
 
 
 def _is_complete(sql):
-    # A complete command is an sql statement that ends with a semicolon, unless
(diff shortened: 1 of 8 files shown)


mssqlcli/mssqlcliclient.py
```diff
@@ -277,7 +278,8 @@ def _exception_found_in(query_response):
 
     @staticmethod
     def _no_results_found_in(query_response):
-        return not query_response.batch_summaries[0].result_set_summaries
+        return not query_response.batch_summaries \
+               or not query_response.batch_summaries[0].result_set_summaries
```
ellbosch: The changes in this file fix a bug that causes multi-line statements to crash when multiple statements are batched (i.e. multiple `select` statements in one multi-line query).

mssqlcli/packages/parseutils/utils.py
```diff
@@ -113,7 +113,7 @@ def is_open_quote(sql):
 
 def _parsed_is_open_quote(parsed):
     # Look for unmatched single quotes, or unmatched dollar sign quotes
-    return any(tok.match(Token.Error, ("'", "$")) for tok in parsed.flatten())
+    return any(tok.match(Token.Error, ("'", '"', "$")) for tok in parsed.flatten())
```
ellbosch: This change adds support for double quotes. The purpose of this method is to catch statements with unclosed quotations.

tests/test_multiline.py
```diff
@@ -0,0 +1,26 @@
...
+        ('select 1; GO', False),
+        ('select 1\n select 2;\ngo', True),
+        ('select 1;', False),
+        ('select 1 go', False)
+        # tests below to be enabled when sqlparse supports retaining newlines
+        # when stripping comments:
+        # ('select 3 /* another open comment\n*/   GO', True),
+        # ('select 1\n*/go', False),
+        # ('select 1 /*\nmultiple lines!\n*/go', True)
+    ]
```
ellbosch: The commented-out tests showcase examples of tests currently lacking support.

tests/test_multiline.py
```diff
@@ -0,0 +1,26 @@
...
+from mssqlcli.mssqlbuffer import _is_complete
+
+
+class TestMssqlCliMultiline:
+    testdata = [
+        (None, False),
+        ('', False),
+        ('select 1 /* open comment!\ngo', False),
+        ('select 1\ngo -- another comment', True),
+        ('select 1; select 2, "open quote: go', False),
```
udeeshagautam: go within quotes - "go"

tests/test_multiline.py
```diff
@@ -0,0 +1,26 @@
...
+        ('select 1; GO', False),
```
udeeshagautam: Caps fails? Sorry if you already explained and I forgot

tests/test_multiline.py
```diff
@@ -0,0 +1,26 @@
...
+        ('select 1\n select 2;\ngo', True),
+        ('select 1;', False),
+        ('select 1 go', False)
```
udeeshagautam: okay what happens when the like has more than on go - go go

mssqlcli/mssqlbuffer.py
```diff
@@ -21,10 +22,28 @@ def cond():
...
+
+        # check that 'go' is only token on newline
+        lines = sql.split('\n')
+        lastline = lines[len(lines) - 1].lower().strip()
+        is_valid_go_on_lastline = 'go' == lastline
+
+        # check that 'go' is on last line, not in open quotes, and there's no open
+        # comment with closed comments and quotes removed
+        return not is_open_quote(sql) and not is_open_comment and is_valid_go_on_lastline
+
```
udeeshagautam: add a comment for why the newline comments dont work and a link to sqlparse bug also for future ref

tests/test_multiline.py
ellbosch: Interestingly, for sqlcmd it fails with: `Sqlcmd: Error: Syntax error at line 2 near command 'go'.`

I added a test case which successfully fails.

tests/test_multiline.py
```diff
@@ -0,0 +1,29 @@
...
+        ('select 1 go', False),
+        ('select 1\ngo go go', False)
+        # tests below to be enabled when sqlparse supports retaining newlines
+        # when stripping comments (tracking here: https://github.com/andialbrecht/sqlparse/issues/484):
+        # ('select 3 /* another open comment\n*/   GO', True),
+        # ('select 1\n*/go', False),
+        # ('select 1 /*\nmultiple lines!\n*/go', True)
+    ]
+
+    @staticmethod
```
chlafreniere: please put some comments here around what this test is doing exactly. someone may look at this 5 years from now and be a little confused what is_complete refers to 😄

tests/test_multiline.py
```diff
@@ -0,0 +1,29 @@
+import pytest
+from mssqlcli.mssqlbuffer import _is_complete
+
+
+class TestMssqlCliMultiline:
+    testdata = [
+        (None, False),
+        ('', False),
+        ('select 1 /* open comment!\ngo', False),
```
chlafreniere: how about some tests where you have a go at the beginning followed by some select statement

PR #400, opened 2019-12-22
Update windows.md

suresh-kandoth:
Lot of folks encounter the python launcher error described in https://stackoverflow.com/questions/24627525/fatal-error-in-launcher-unable-to-create-process-using-c-program-files-x86. It is better to provide an alternate installation path that will not encounter that error.


ellbosch:
Thanks @suresh-kandoth!

Issue #411, opened 2020-02-07
Horizontal paging should be supported out of the box


ellbosch:
Use Less pager if already installed.

johanwk:
Hi, I have this problem on Windows 10 with _less_ not being found. I have less installed with Git for Windows, but mssql-cli responds with "The specified path is invalid.".

Advice on getting this to work by adding a precompiled less.exe -- is there a binary somewhere that you would recommend, that could solve this problem?

ellbosch:
@johanwk we're exploring our options at this moment. You may try finding a compiled solution online of less.exe at your own risk, unfortunately we cannot recommend any solution.

johanwk:
@ellbosch Fully understood -- thank you for the update.

PR #423, opened 2020-02-26
Set auto expand to false if less pager is installed

ellbosch:
This PR brings support for horizontal paging out-of-the-box, given the following criteria are met:
1. less is installed on machine, and
2. vertical output setting is set to auto or never

For Linux users, this should bring horizontal paging as a standard experience, given that less is usually installed by default.

Tag: #411


ellbosch:
@pensivebrian I needed to change our approach slightly since my implementation doesn't work on Python 2.

I'll look at tests now. It may be imperfect since we'll be testing without a less pager installed on these machines.

ellbosch:
@pensivebrian bugs are fixed and the PR is ready for re-review again :)

PR #425, opened 2020-02-28
Fixed bug which caused some queries with semicolons to crash

### LINKED ISSUE #422: app fails to execute merge statement and reports "A MERGE statement must be terminated by a semi-colon (;)."
Version: 0.18.0
Under bash, using the attached sql file, exec 
`mssql-cli -U $user -d $database -S $host -i test-basic-merge.sql`

Get error "A MERGE statement must be terminated by a semi-colon (;)."

This file properly executes using another client, DataGrip.

[test-basic-merge.sql.txt](https://github.com/dbcli/mssql-cli/files/4248168/test-basic-merge.sql.txt)

ellbosch:
Fixes #422 

The mssql-cli client uses old and unnecessary logic to right-strip semicolons. This was introduced in a refactor a few years ago, making it unclear why it was added in the first place. All of our tests passed after removing the semicolon-strip logic, however we needed to make minor updates to a few tests to expect semicolons in the input text.

This PR also introduces new tests to ensure the script in the #422 bug report work in multiline and non-interactive modes.

### DIFF
--- a/tests/test_mssqlcliclient.py
+++ b/tests/test_mssqlcliclient.py
@@ -240,7 +240,7 @@ class TestMssqlCliClientMultipleStatement(MssqlCliClient):
         for (i, (rows, _, _, query_executed, is_error)) in \
             enumerate(client.execute_query(query_str)):
 
-            queries = query_str.split(';')
+            queries = ["{};".format(query) for query in query_str.split(';')]
             assert query_executed == queries[i].strip()
             assert len(rows) == rows_outputted[i]
             assert (is_error and len(rows) == 0) or (not is_error)
--- a/tests/test_mssqlcliclient.py
+++ b/tests/test_mssqlcliclient.py
@@ -12,7 +12,9 @@ from mssqltestutils import (
     create_test_db,
     clean_up_test_db,
     shutdown,
-    random_str
+    random_str,
+    get_io_paths,
+    get_file_contents
 )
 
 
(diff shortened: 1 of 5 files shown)


tests/test_mssqlcliclient.py
```diff
@@ -240,7 +242,20 @@ def test_mssqlcliclient_multiple_statement(client, query_str, rows_outputted):
         for (i, (rows, _, _, query_executed, is_error)) in \
             enumerate(client.execute_query(query_str)):
 
-            queries = query_str.split(';')
+            queries = ["{};".format(query) for query in query_str.split(';')]
```
pensivebrian: If a literal string has a semi-colon would this parse? Looks like an existing bug so this doesn't block this change. Can we add a test for this an confirm, and file a bug if needed?

tests/test_query_inputs/input_multiple_merge.txt
```diff
@@ -0,0 +1,62 @@
+--Example taken from https://www.sqlservertutorial.net/sql-server-basics/sql-server-merge/
```
pensivebrian: let's remove this comment

tests/test_mssqlcliclient.py
ellbosch: This is just our testing functionality—semicolons parse just fine in the product code. We could hard-code the array if we want to add a test with semicolons.

pensivebrian: oh this is just a test - my bad  :)

pensivebrian: However, are we testing literal strings :)  :)

PR #426, opened 2020-02-29
Added multiline tests

ellbosch:


tests/test_multiline.py
```diff
@@ -17,7 +17,9 @@ class TestMssqlCliMultiline:
         ('select 1 go', False),
         ('select 1\ngo go go', False),
         ('GO select 1', False),
-        ('GO', True)
+        ('GO', True),
+        ('select 1;select 1\nGO\nselect 1;select 1\nGO', True),
```
pensivebrian: Can we add named parameters for readability? `is_complete: true `or whatever the syntax is. Also, what is is_complete is a little ambiguous. Perhaps a better name: is_valid_sqlcmd?

ellbosch: Sorry for getting to this so late—I changed the method name from `is_complete` to `is_query_executable`. I think this is an improvement because it more clearly designates that the multiline query would be processable. Thoughts?

pensivebrian: yes, that's much more readable

PR #431, opened 2020-03-04
Created automation for Packages.Microsoft publishing

ellbosch:
In the spirt of our deb and rpm packaging, I have created a Dockerized approach to automate* the publishing of deb and rpm packages to the Packages.Microsoft repository. Automation is needed to support the very long list of Linux distributions we've previously supported. **A future PR will provide improved testing for our supported Linux distributions**.

\* _The Repo CLI used to publish packages requires manual installation. This solution is not perfectly automated due to this complexity. Therefore, a step-by-step guide is provided._

Details of this PR:
- **Dockerfile which creates container for publishing**. Packages.Microsoft provides a Repo CLI API for publishing packages, which requires an AMD64 environment. We use a Dockerfile to create the staging environment needed for Repo CLI installation and script automation.
- **publish.sh script uploads to either testing or prod.** When called in the pre-built Docker container, this script will publish a deb or rpm to each supported package in our json files. _It requires an argument for specifying either 'testing' or 'prod', and will not execute unless specified._
- **JSON files for testing and prod-supported repos**. The objects refers to identifiers known by Packages.Microsoft, which are pulled during `publish.sh` execution. Each entity represents a Linux distribution where we've previously published mssql-cli, _or_ its future versions we've yet to support.
- **Step-by-step docs**. A README is provided to get the Docker container up and running, to install the Repo CLI, and to execute the `publish.sh` script.
- **Folder organization changes**. A new 'release_scripts' folder has been added to the repo root, where automated tasks for publishing may reside (currently hosts a 'Packages.Microsoft' folder, only).


release_scripts/Packages.Microsoft/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+# Add mssql-cli repo
+WORKDIR /root
+RUN mkdir Repos
+RUN mkdir Repos/mssql-cli
+ADD . Repos/mssql-cli
+
+# Download latest stable deb and rpm packages
+WORKDIR /root
+RUN wget https://mssqlcli.blob.core.windows.net/daily/deb/mssql-cli-dev-latest.deb
+RUN wget https://mssqlcli.blob.core.windows.net/daily/rpm/mssql-cli-dev-latest.rpm
```
pensivebrian: Let's move away from the 'latest' package and use explicit version numbers. This issue is, a later build might have overwritten the latest and we would know. If we reference the explicit version, this is less likely to happen.

release_scripts/Packages.Microsoft/supported_repos_testing.json
```diff
@@ -0,0 +1,280 @@
+[
```
pensivebrian: Is this coming from a package? If so, can we just download package and avoid checking this is? My concern is these values can change.

release_scripts/Packages.Microsoft/supported_repos_prod.json
```diff
@@ -0,0 +1,126 @@
+
```
pensivebrian: same deal here - do we have to check this in?

release_scripts/Packages.Microsoft/supported_repos_testing.json
ellbosch: These values won't change—they map to repos in Packages.Microsoft. If you query the Repo CLI you can get this information.

The key bit of data is the repository ID, which needs to be in the root config when uploading a package. Without some sort of automation, you'd need to look up the repository ID for each Linux repo you're uploading to, and edit the root config manually.

release_scripts/Packages.Microsoft/supported_repos_prod.json
ellbosch: See my comment above. We'd need to store this data somewhere for automation to work. There isn't sensitive data so I thought a JSON would be ok.

As an alternative solution, we could store these values in a database.

release_scripts/Packages.Microsoft/publish.sh
```diff
@@ -0,0 +1,53 @@
...
+elif [ ${2,,} = 'testing' ]; then
+    repo_dist='testing'
+else
+    echo "Second argument should specify 'prod' or 'testing' for repository distribution type."
+    exit 1
+fi
+
+local_repo=$1
+config_file=/root/.repoclient/config.json
+deb_pkg=/root/mssql-cli-dev-latest.deb
```
ellbosch, line 20: remove dev-latest

release_scripts/Packages.Microsoft/publish.sh
```diff
@@ -0,0 +1,53 @@
...
+    repo_url=$(_jq '.url')
+
+    # make temp config with new id value and then replace repoclient config
+    jq --arg a "$repo_id" '.repositoryId = $a' $config_file > $tmp_dir/tmp_config.json \
+        && mv $tmp_dir/tmp_config.json $config_file
+
+    # with config updated, publish deb or rpm package
+    if [ $repo_type == "apt" ]; then
+        echo "Publishing .deb for $repo_url..."
+        repoclient package add $deb_pkg
```
ellbosch, line 46: add to list of commands in readme

release_scripts/Packages.Microsoft/supported_repos_prod.json
ellbosch: remove json, keep urls and add to script

release_scripts/Packages.Microsoft/publish.sh
```diff
@@ -0,0 +1,100 @@
...
+        echo ${repo_data} | base64 --decode | jq -r ${1}
+    }
+    repo_id=$(_jq '.id')
+    repo_type=$(_jq '.type')
+
+    # publish deb or rpm package
+    # -r specifies the destination repository (by ID)
+    if [ $repo_type == "apt" ]; then
+        echo "Publishing .deb for $repo_url..."
+        repoclient package add $deb_pkg -r $repo_id
```
pensivebrian, line 93: what if there is an error? If there is an error, I think we should abort.

release_scripts/Packages.Microsoft/Dockerfile
```diff
@@ -0,0 +1,31 @@
+FROM amd64/ubuntu:18.04 AS builder
```
pensivebrian, line 1: You can mentioned there are issues with this. Does this work out of the box?

release_scripts/Packages.Microsoft/README.md
```diff
@@ -0,0 +1,89 @@
...
+```
+
+Visit the [testing](#testing-distribution-downloads) section for more information on downloading builds from the testing distribution.
+
+### Publish to Production
+When ready for production, change `'testing'` with `'prod'`, which will upload to all repositories in the `supported_repos_prod.json` file:
+```sh
+release_scripts/Packages.Microsoft/publish.sh $(pwd) 'prod'
+```
+
```
pensivebrian, line 61: We should add a note about what todo if there is an error or need to refresh the binaries with a new version. Can we just re-run the tool with the same package versions?

release_scripts/Packages.Microsoft/publish.sh
ellbosch: Great minds think alike!!! I had problems terminating the script during this step so I just made a change :)

release_scripts/Packages.Microsoft/publish.sh
```diff
@@ -0,0 +1,100 @@
+# Publishes deb and rpm packages to Packages.Microsoft repo
```
pensivebrian, line 1: I'm curious why we didn't do this in python?

release_scripts/Packages.Microsoft/Dockerfile
ellbosch: The only issue I remember bringing up was the requirement for using amd64 to install the Repo Client. So with Docker, this should work out of the box.

release_scripts/Packages.Microsoft/publish.sh
```diff
@@ -0,0 +1,100 @@
...
+fi
+
+local_repo=$1
+deb_pkg=/root/mssql-cli_*.deb
+rpm_pkg=/root/mssql-cli-*.rpm
+
+# specificies destination testing repos for publishing
+repo_url_testing=( \
+    "microsoft-ubuntu-trusty-prod" \
+    "microsoft-ubuntu-xenial-prod" \
```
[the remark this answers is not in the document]
ellbosch: Nope—the URL naming is unfortunately confusing and inconsistent at times.

In some cases, the URL name is the same for both the prod and testing distribution. However, repo ID's are always unique, and we query for the desired repo ID by filtering for URL name *and* distribution type (i.e. prod or testing).

release_scripts/Packages.Microsoft/Dockerfile
pensivebrian: I thought you mention stdio was involved for a license agreement or something.

release_scripts/Packages.Microsoft/publish.sh
ellbosch: Looking back in hindsight... we could have. Pros and cons here. I thought the bash script would have been more elegant because I pictured fewer complications than our result.

What's nice here is you don't need any Python, it works in any container with bash so this script can theoretically be used in any Linux container (using arm). But Python would have been a cleaner looking script.

release_scripts/Packages.Microsoft/README.md
ellbosch: Do you mean package versions of mssql-cli? Yes.

release_scripts/Packages.Microsoft/Dockerfile
ellbosch: Oh, yes, STDIO is required for responding to a prompt with the repo client installer. That's why the repo client installation is out of the dockerfile's scope—instead, the dockerfile completes requirements needed for repo client installation.

release_scripts/Packages.Microsoft/README.md
pensivebrian: I'm not sure what I mean. I just remember you mentioned there was an issue with stdio and the repro tool. If there is an issue, can you refresh my memory?

pensivebrian: ops, wrong thread. Cool, we should doc this since many repo's (nuget, pypi) don't support this (or didn't in the recent past)

ellbosch: Let me run this back with the Repo Client team to confirm.

release_scripts/Packages.Microsoft/Dockerfile
pensivebrian: I see. Can we have a similar script for those who don't use docker?

ellbosch: If you don't use Docker, then it's just a matter of building a Linux machine that can install the repo client. If that's done, then the only remaining steps are:
1. Install Repo Client
2. Call the publish.sh file

No other scripts are needed.

PR #435, opened 2020-03-06
Added support for Linux package codesigning

ellbosch:
> Note: these changes have already been validated with Azure DevOps with successful runs.

This PR updates our Dockerfile scripts for debian and rpm packaging in the following ways:
* **Adds support for codesigning.** Dockerfiles have new commands to copy deb and rpm packages to the root of the container, which enables Docker to copy packages from the container to the host. With packages now in the host, we can run the ESRP DevOps codesign task.
* **Removes Python dependency installation in Dockerfile**. DevOps pipelines have been updated to publish packages directly from the host, rather than from Docker. With this change, there's no longer a dependency on our Python code to publish deb and rpm packages to daily storage.
* **Updated docs.** Release docs have now been updated to reflect the aforementioned changes to publishing. I've also added the `sh` suffix to codeblocks for formatting.


build.py
```diff
@@ -80,10 +80,13 @@ def build():
...
-        utility.exec_command('twine check {}'
-                             .format(os.path.join(utility.MSSQLCLI_DIST_DIRECTORY, '*')),
-                             utility.ROOT_DIR, continue_on_error=False)
+        try:
+            # checks if long description will render correctly--does not work on some systems
+            utility.exec_command('twine check {}'
+                                .format(os.path.join(utility.MSSQLCLI_DIST_DIRECTORY, '*')),
+                                utility.ROOT_DIR, continue_on_error=False)
+        except IOError:
+            print("Unable to run 'twine check'.")
```
pensivebrian: Let's print the exception error message and stack trace so that someone could troubleshoot

build_scripts/debian/Dockerfile
```diff
@@ -16,11 +16,12 @@ RUN apt-get -y install python3-all python3-pip python3-setuptools nano git sudo
 RUN python3 -m pip install --upgrade pip
 
 # Build .deb
-RUN mkdir Repos
-RUN mkdir Repos/mssql-cli
-ADD . Repos/mssql-cli
-WORKDIR Repos/mssql-cli
+RUN mkdir /Repos
```
pensivebrian: So we're hardcoding and assuming a /Repos/mssql-cli? Seems fine for docker, but we can't make this assumption for dev boxes.

ellbosch: For sure, I've accounted for this.

PR #436, opened 2020-03-08
Updated stable download links

ellbosch:
I've updated links to point at latest stable version number, rather than use 'dev-latest'.

Furthermore, all downloads now point to our daily storage. @pensivebrian please let me know if this is good practice.

### DIFF
--- a/doc/installation/README.md
+++ b/doc/installation/README.md
@@ -16,10 +16,10 @@ also be found in the links above.
 
 | Supported Platform                         |Latest Stable                 |
 |--------------------------------------------|------------------------------|
-|  Windows (x64)                             |[.whl][whl-win-x64]           |
-|  Windows (x86)                             |[.whl][whl-win-x86]           |
-|  macOS 10.12+                              |[.whl][whl-macos]             |
-|  Linux (Python Wheel)                      |[.whl][whl-linux]             |
+|  Windows (x64)                             |[.whl][whl]           |
+|  Windows (x86)                             |[.whl][whl]           |
+|  macOS 10.12+                              |[.whl][whl]             |
+|  Linux (Python Wheel)                      |[.whl][whl]             |
 |  Ubuntu 14.04+                             |[.deb][deb]                   |
 |  Debian 8.7+                               |[.deb][deb]                   |
 |  CentOS 7+                                 |[.rpm][rpm]                   |
@@ -28,14 +28,8 @@ also be found in the links above.
 |  Fedora 25+                                |[.rpm][rpm]                   |
 
 
-[deb]: https://mssqlcli.blob.core.windows.net/daily/deb/mssql-cli_0.18.1-1_all.deb
+[deb]: https://packages.microsoft.com/ubuntu/16.04/prod/pool/main/m/mssql-cli/mssql-cli_1.0.0-1_all.deb
 
-[rpm]: https://mssqlcli.blob.core.windows.net/daily/rpm/mssql-cli-0.18.1-1.el8.x86_64.rpm
+[rpm]: https://packages.microsoft.com/centos/7/prod/mssql-cli-1.0.0-1.el7.x86_64.rpm
 
-[whl-win-x64]: https://mssqlcli.blob.core.windows.net/daily/whl/mssql-cli/mssql_cli-0.18.1-py2.py3-none-win_amd64.whl
-
-[whl-win-x86]: https://mssqlcli.blob.core.windows.net/daily/whl/mssql-cli/mssql_cli-0.18.1-py2.py3-none-win32.whl
-
-[whl-macos]: https://mssqlcli.blob.core.windows.net/daily/whl/mssql-cli/mssql_cli-0.18.1-py2.py3-none-macosx_10_11_intel.whl
-
-[whl-linux]: https://mssqlcli.blob.core.windows.net/daily/whl/mssql-cli/mssql_cli-0.18.1-py2.py3-none-manylinux1_x86_64.whl
+[whl]: https://pypi.org/manage/project/mssql-cli/release/1.0.0/
--- a/doc/installation/README.md
+++ b/doc/installation/README.md
@@ -16,10 +16,10 @@ also be found in the links above.
 
 | Supported Platform                         |Latest Stable                 |
 |--------------------------------------------|------------------------------|
-|  Windows (x64)                             |[.whl][whl]           |
-|  Windows (x86)                             |[.whl][whl]           |
-|  macOS 10.12+                              |[.whl][whl]             |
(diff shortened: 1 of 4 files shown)


ellbosch:
> Stable bits should never point to our storage account - we should link to the official repo.

Unfortunately with regards to PyPI, there's a unique link that's generated for each wheel that's upload. You can see this [here](https://pypi.org/manage/project/mssql-cli/release/0.18.1/). What do you think is the best way to manage direct links to wheels?

doc/installation/README.md
```diff
@@ -19,15 +19,23 @@ also be found in the links above.
...
-[rpm]: https://packages.microsoft.com/rhel/7/prod/mssql-cli-0.15.0-1.el7.x86_64.rpm
-[whl-win-x64]: https://files.pythonhosted.org/packages/79/31/1be42f3632a30bb126e02cda4312b797f5f2cdef60b9c62596196a475037/mssql_cli-0.15.0-py2.py3-none-win_amd64.whl
-[whl-win-x86]: https://files.pythonhosted.org/packages/21/8c/9829c2094b4f179f9c2f4548b2ba089867f84a3758cb70575ef2a905e877/mssql_cli-0.15.0-py2.py3-none-win32.whl
-[whl-macos]: https://files.pythonhosted.org/packages/43/5d/c9af6aec5b491e7b0c5ccf00b4b8062282d6c4cfb4c0417891bd6013e299/mssql_cli-0.15.0-py2.py3-none-macosx_10_11_intel.whl
+
+[deb]: https://mssqlcli.blob.core.windows.net/daily/deb/mssql-cli_0.18.1-1_all.deb
+
+[rpm]: https://mssqlcli.blob.core.windows.net/daily/rpm/mssql-cli-0.18.1-1.el8.x86_64.rpm
+
+[whl-win-x64]: https://mssqlcli.blob.core.windows.net/daily/whl/mssql-cli/mssql_cli-0.18.1-py2.py3-none-win_amd64.whl
```
pensivebrian: We shouldn't be pointing people at our blob storage account. Can you update these to point them to pypi? We should see if there is a way to gen the link as part of the version bump, though it becomes more complicated due to the hash.

doc/installation/README.md
```diff
@@ -19,15 +19,23 @@ also be found in the links above.
...
 |  openSUSE                                  |[.rpm][rpm]                   |
 |  Fedora 25+                                |[.rpm][rpm]                   |
 
-[deb]: https://packages.microsoft.com/ubuntu/14.04/prod/pool/main/m/mssql-cli/mssql-cli_0.15.0-1_all.deb
-[rpm]: https://packages.microsoft.com/rhel/7/prod/mssql-cli-0.15.0-1.el7.x86_64.rpm
-[whl-win-x64]: https://files.pythonhosted.org/packages/79/31/1be42f3632a30bb126e02cda4312b797f5f2cdef60b9c62596196a475037/mssql_cli-0.15.0-py2.py3-none-win_amd64.whl
-[whl-win-x86]: https://files.pythonhosted.org/packages/21/8c/9829c2094b4f179f9c2f4548b2ba089867f84a3758cb70575ef2a905e877/mssql_cli-0.15.0-py2.py3-none-win32.whl
-[whl-macos]: https://files.pythonhosted.org/packages/43/5d/c9af6aec5b491e7b0c5ccf00b4b8062282d6c4cfb4c0417891bd6013e299/mssql_cli-0.15.0-py2.py3-none-macosx_10_11_intel.whl
+
+[deb]: https://mssqlcli.blob.core.windows.net/daily/deb/mssql-cli_0.18.1-1_all.deb
```
pensivebrian: Same deal for the .deb and .rpm packages - they should point to packages.microsoft.com.

doc/installation/README.md
ellbosch: See my above comment on PyPI. Not sure what to do there. I'll get to RPM and DEB, however.

ellbosch:
> Need to point links to pypi and packages.microsoft.com.

Here's a thought. What if we just removed all direct links? I just don't see the need to provide this, and also given my comments above on not being able to provide direct links to PyPI.

doc/installation/README.md
```diff
@@ -16,18 +16,20 @@ also be found in the links above.
...
 |  openSUSE                                  |[.rpm][rpm]                   |
 |  Fedora 25+                                |[.rpm][rpm]                   |
 
-[deb]: https://packages.microsoft.com/ubuntu/14.04/prod/pool/main/m/mssql-cli/mssql-cli_0.15.0-1_all.deb
-[rpm]: https://packages.microsoft.com/rhel/7/prod/mssql-cli-0.15.0-1.el7.x86_64.rpm
-[whl-win-x64]: https://files.pythonhosted.org/packages/79/31/1be42f3632a30bb126e02cda4312b797f5f2cdef60b9c62596196a475037/mssql_cli-0.15.0-py2.py3-none-win_amd64.whl
-[whl-win-x86]: https://files.pythonhosted.org/packages/21/8c/9829c2094b4f179f9c2f4548b2ba089867f84a3758cb70575ef2a905e877/mssql_cli-0.15.0-py2.py3-none-win32.whl
-[whl-macos]: https://files.pythonhosted.org/packages/43/5d/c9af6aec5b491e7b0c5ccf00b4b8062282d6c4cfb4c0417891bd6013e299/mssql_cli-0.15.0-py2.py3-none-macosx_10_11_intel.whl
+
+[deb]: https://packages.microsoft.com/ubuntu/16.04/prod/pool/main/m/mssql-cli/mssql-cli_1.0.0-1_all.deb
```
pensivebrian, line 31: One issue is these files don't exist yet until we publish, so when we merge the vbump into master, the links won't work. I'm leaning towards just hand editing the links after we push to pypi and the other repos.

ellbosch:
Holding off on this until 1.0 is released.

ellbosch:
@pensivebrian: my most recent commits update the Linux docs:
* Added Ubuntu 18.04 and CentOS 8 instructions
* Removed SUSE and Fedora instructions
* Added import key step for RPM builds so that the signature warning no longer appears

ellbosch:
> For the install steps, we have the same steps for the other [SQL Linux tools](https://docs.microsoft.com/en-us/sql/linux/sql-server-linux-setup-tools?view=sql-server-ver15) which have had a fair amount of testing. I'd like to keep them the same unless there is good reason to diverge.

I had chosen to go off the docs for [Packages.Microsoft](https://docs.microsoft.com/en-us/windows-server/administration/linux-package-repository-for-microsoft-software), but I can change if you prefer SQL Server docs.

RPM instructions appear to be the only difference. SQL Server docs are missing a step to import the key, without which leads to signature warning, so I'll make sure to keep that.

PR #441, opened 2020-03-24
Stabilized db create for tests

### LINKED ISSUE #280: Create database against Azure needs to more resilient.

### LINKED ISSUE #428: Stop using permanent DB for testing
We're currently using a permanent DB for testing. We need to create temporary DBs when tests run.

Also: use our old SQL server instance.

ellbosch:
Closes #280.

This PR introduces retry logic to check the status of a db creation using `sys.dm_operation_status`.

### DIFF
--- a/tests/mssqltestutils.py
+++ b/tests/mssqltestutils.py
@@ -1,6 +1,6 @@
+from datetime import datetime, timedelta
 import os
 import socket
-import time
 import warnings
 from argparse import Namespace
 import mssqlcli.sqltoolsclient as sqltoolsclient
@@ -94,7 +94,7 @@ def create_test_db():
     """
     Creates database for test, using various status checks and retry logic for reliability.
     - Calls helper method to check status of create db task, if possible
-    - Exits on successful response or retry counter exceeds limit
+    - Exits on successful response or retry period exceeds time limit
     """
     client = create_mssql_cli_client()
     local_machine_name = socket.gethostname().replace(
@@ -135,17 +135,15 @@ def check_create_db_status(db_name, client):
     query_check_status = u"SELECT TOP 1 state_desc FROM sys.dm_operation_status " \
                          u"WHERE major_resource_id = '{}' AND operation = 'CREATE DATABASE' " \
                          u"ORDER BY start_time DESC".format(db_name)
-    count = 0
-    while count < 5:
+
+    # retry for 5 minutes until db status is no longer 'processing'
+    datetime_end_loop = datetime.now() + timedelta(minutes=5)
+    while datetime.now() < datetime_end_loop:
         for row, _, _, _, _ in client.execute_query(query_check_status):
             create_db_status = row[0][0]
 
             if create_db_status != 'PROCESSING':
                 return create_db_status
-
-            # sleep for 5 seconds and try again if still processing
-            time.sleep(5)
-            count += 1
 
     return 'FAILED'
 
--- a/tests/mssqltestutils.py
+++ b/tests/mssqltestutils.py
@@ -93,61 +93,65 @@ def getTempPath(*args):
 def create_test_db():
     """
     Creates database for test, using various status checks and retry logic for reliability.
-    - Uses outer while loop to retry db create failures
-    - Uses an inner while loop to keep checking status if still processing
-    - Exits on successful response
+    - Uses while loop to retry db create failures
+    - Calls helper method to check status of create db task, if possible
+    - Exits on successful response or outer/inner counters exceed limit
     """
     client = create_mssql_cli_client()
     local_machine_name = socket.gethostname().replace(
         '-', '_').replace('.', '_')
 
     # retry logic in case db create fails
-    count_outer = 0
-    while count_outer < 5:
+    count = 0
+    while count < 5:
         test_db_name = u'mssqlcli_testdb_{0}_{1}'.format(
             local_machine_name, random_str())
         query_db_create = u"CREATE DATABASE {0};".format(test_db_name)
 
         for _, _, status, _, is_create_error in client.execute_query(query_db_create):
-            if is_create_error:
+            create_db_status = check_create_db_status(test_db_name, client)
+
+            if is_create_error or create_db_status == 'FAILED':
                 # log warning to console and cleanup db
                 warnings.warn('Test DB create failed with error: {0}'.format(status))
                 clean_up_test_db(test_db_name)
-            else:
-                # check if create db is still processing. if so, continue to check on it for
-                # another 5 tries
-                count_inner = 0
-                while count_inner < 5:
-                    create_db_status = check_create_test_db_status(test_db_name, client)
-                    if create_db_status == 'COMPLETED':
-                        # created successfully
-                        shutdown(client)
-                        return test_db_name
 
-                    if create_db_status == 'PROCESSING':
-                        # sleep for 5 seconds and try again
-                        time.sleep(5)
-                        count_inner += 1
-                    else:
-                        # db create failed, cleanup db, and exit inner while loop
-                        warnings.warn('Test DB create failed. Evaluated from ' \
-                                      'sys.dm_operation_status.')
-                        clean_up_test_db(test_db_name)
-                        break
-        count_outer += 1
+            elif create_db_status in ('COMPLETED', 'UNKNOWN'):
+                shutdown(client)
+                return test_db_name
+
+        count += 1
+
     shutdown(client)
 
     # cleanup db just in case, then raise exception
     clean_up_test_db(test_db_name)
     raise AssertionError("DB creation failed.")
 
-def check_create_test_db_status(db_name, client):
-    """ Checks if database create is still being processed. """
+def check_create_db_status(db_name, client):
+    """
+    Uses retry logic with sys.dm_operation_status to check statis of create database job.
+    """
+    if client.database.lower() != 'master':
+        # sys.dm_operation_status CANNOT be used if the client isn't running on master db.
+        return 'UNKNOWN'
+
     query_check_status = u"SELECT TOP 1 state_desc FROM sys.dm_operation_status " \
                          u"WHERE major_resource_id = '{}' AND operation = 'CREATE DATABASE' " \
                          u"ORDER BY start_time DESC".format(db_name)
-    for row, _, _, _, _ in client.execute_query(query_check_status):
-        return row[0][0]
+    count = 0
+    while count < 5:
+        for row, _, _, _, _ in client.execute_query(query_check_status):
+            create_db_status = row[0][0]
+
+            if create_db_status != 'PROCESSING':
+                return create_db_status
+
+            # sleep for 5 seconds and try again if still processing
+            time.sleep(5)
+            count += 1
+
+    return 'FAILED'
 
 def clean_up_test_db(test_db_name):
     client = create_mssql_cli_client()
--- a/tests/mssqltestutils.py
+++ b/tests/mssqltestutils.py
@@ -93,34 +93,30 @@ def getTempPath(*args):
 def create_test_db():
     """
     Creates database for test, using various status checks and retry logic for reliability.
-    - Uses while loop to retry db create failures
     - Calls helper method to check status of create db task, if possible
-    - Exits on successful response or outer/inner counters exceed limit
+    - Exits on successful response or retry counter exceeds limit
     """
     client = create_mssql_cli_client()
     local_machine_name = socket.gethostname().replace(
         '-', '_').replace('.', '_')
 
-    # retry logic in case db create fails
-    count = 0
-    while count < 5:
-        test_db_name = u'mssqlcli_testdb_{0}_{1}'.format(
-            local_machine_name, random_str())
-        query_db_create = u"CREATE DATABASE {0};".format(test_db_name)
+    test_db_name = u'mssqlcli_testdb_{0}_{1}'.format(
+        local_machine_name, random_str())
+    query_db_create = u"CREATE DATABASE {0};".format(test_db_name)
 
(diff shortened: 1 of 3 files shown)


tests/mssqltestutils.py
```diff
@@ -90,29 +91,68 @@ def getTempPath(*args):
...
+    """
+    if client.database.lower() != 'master':
+        # sys.dm_operation_status CANNOT be used if the client isn't running on master db.
+        return 'UNKNOWN'
+
+    query_check_status = u"SELECT TOP 1 state_desc FROM sys.dm_operation_status " \
+                         u"WHERE major_resource_id = '{}' AND operation = 'CREATE DATABASE' " \
+                         u"ORDER BY start_time DESC".format(db_name)
+    count = 0
+    while count < 5:
```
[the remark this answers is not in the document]
ellbosch: Not positive the best way to do this, but perhaps we change line 151 to sleep for 60 seconds instead of 5. Thoughts?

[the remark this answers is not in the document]
pensivebrian: that works, but a brittle approach. a better way: grab the current and add five minutes which become the timeout. then loop until the current time is greater than the timeout.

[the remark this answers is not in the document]
ellbosch: I like that approach a lot... ok will update.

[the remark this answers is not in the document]
ellbosch: Updated. Let me know what you think, thanks Brian!

tests/mssqltestutils.py
```diff
@@ -90,29 +91,62 @@ def getTempPath(*args):
...
     clean_up_test_db(test_db_name)
     raise AssertionError("DB creation failed.")
 
+def check_create_db_status(db_name, client):
+    """
+    Uses retry logic with sys.dm_operation_status to check statis of create database job.
+    """
+    if client.database.lower() != 'master':
+        # sys.dm_operation_status CANNOT be used if the client isn't running on master db.
+        return 'UNKNOWN'
```
pensivebrian, line 133: Probably should throw and error here as it's probably a coding mistake.

ellbosch: Changed functionality here to raise an exception. Create db method now catches the raised exception instead of checking for `'UNKNOWN'`.

tests/mssqltestutils.py
```diff
@@ -90,29 +92,71 @@ def getTempPath(*args):
...
 
     # cleanup db just in case, then raise exception
     clean_up_test_db(test_db_name)
     raise AssertionError("DB creation failed.")
 
+def check_create_db_status(db_name, client):
+    """
+    Uses retry logic with sys.dm_operation_status to check statis of create database job.
+    """
+    if client.database.lower() != 'master':
```
pensivebrian, line 138: This looks good, but can we move this to a higher level? Seems like this check should occur before we issue the 'create database statement'.

ellbosch:
@pensivebrian I've made the suggested changes, please let me know your thoughts.

We still need to fix the known issue in the Special module but I can submit that in a separate PR.

tests/mssqltestutils.py
```diff
@@ -90,29 +92,78 @@ def getTempPath(*args):
...
-    client = create_mssql_cli_client()
+    """
+    Creates database for test, using various status checks and retry logic for reliability.
+    - Calls helper method to check status of create db task, if possible
+    - Exits on successful response or retry period exceeds time limit
+    """
+    options = create_mssql_cli_options(database='master')
+    client = create_mssql_cli_client(options)
+
+    if not _is_client_db_on_cloud(client):
```
pensivebrian, line 103: We should be able to run on-premise. Rather than raising an exception, can you just skip the rety logic for on-premise?

tests/mssqltestutils.py
```diff
@@ -90,29 +92,79 @@ def getTempPath(*args):
...
-        for _, _, status, _, is_create_error in client.execute_query(query_db_create):
-            if not is_create_error:
-                shutdown(client)
-                return test_db_name
-            # log warning to console
+
+    for _, _, status, _, is_create_error in client.execute_query(query_db_create):
+        if is_create_error:
+            # break loop to assert db creation failure
             warnings.warn('Test DB create failed with error: {0}'.format(status))
```
pensivebrian, line 113: we should throw an error or fail the test with the create database exception since there is no point in proceeding.

ellbosch: This still happens—by calling `break` we exit the loop, which raises an assertion failure.

tests/mssqltestutils.py
```diff
@@ -90,29 +92,79 @@ def getTempPath(*args):
...
+                # break loop to assert db creation failure
+                break
+
+        shutdown(client)
+        return test_db_name
+
     shutdown(client)
 
     # cleanup db just in case, then raise exception
     clean_up_test_db(test_db_name)
```
pensivebrian, line 130: >clean_up_test_db [](start = 4, length = 16)

This will most likely fail, so let's wrap in a try catch

tests/mssqltestutils.py
```diff
@@ -90,29 +92,79 @@ def getTempPath(*args):
...
     raise AssertionError("DB creation failed.")
```
pensivebrian, line 131: Can we put the failing exception in the assert message - this way we don't have look at logs

tests/mssqltestutils.py
```diff
@@ -90,29 +91,68 @@ def getTempPath(*args):
...
+        query_db_create = u"CREATE DATABASE {0};".format(test_db_name)
+
         for _, _, status, _, is_create_error in client.execute_query(query_db_create):
-            if not is_create_error:
+            create_db_status = check_create_db_status(test_db_name, client)
+
+            if is_create_error or create_db_status == 'FAILED':
+                # log warning to console and cleanup db
+                warnings.warn('Test DB create failed with error: {0}'.format(status))
+                clean_up_test_db(test_db_name)
```
pensivebrian: >clean_up_test_db [](start = 16, length = 16)

I don't think we should we should try an cleanup here. Let's let the caller perform the cleanup.

tests/mssqltestutils.py
pensivebrian: I see. I missed this.

tests/mssqltestutils.py
```diff
@@ -90,29 +92,79 @@ def getTempPath(*args):
...
+            create_db_status = _check_create_db_status(test_db_name, client)
+
+            if create_db_status == 'FAILED':
+                # break loop to assert db creation failure
+                break
+
+        shutdown(client)
+        return test_db_name
+
     shutdown(client)
```
pensivebrian, line 127: Can we put one shutdown call in a finally block?

tests/test_globalization.py
```diff
@@ -91,161 +121,82 @@ def get_next_characters(all_characters, count=50):
             u'ꆩꆪꆫꆬꆭꆮꆯꆰꆱꆲꆳꆴꆵꆶꆷꆸꆹꆺꆻꆼꆽꆾꆿꇀꇁꇂꇃꇄꇅꇆꇇꇈꇉꇊꇋꇌꇍꇎꇏꇐꇑꇒꇓꇔꇕꇖꇗꇘꇙꇚꇛꇜꇝꇞꇟꇠꇡꇢꇣꇤꇥꇦꇧꇨꇩꇪꇫꇬꇭꇮꇯꇰꇱꇲꇳꇴꇵꇶꇷꇸꇹꇺꇻꇼ'
             u'ꇽꇾꇿꈀꈁꈂꈃꈄꈅꈆꈇꈈꈉ'
         ),
-        'zang': (
+        # zang
```
pensivebrian, line 124: We remove the language key? With the key, you could narrow down the failure by logging the language

tests/test_globalization.py
```diff
@@ -91,161 +121,82 @@ def get_next_characters(all_characters, count=50):
...
-        self.run_charset_validation('tibetian')
-
-    def test_charset_yi(self):
-        self.run_charset_validation('yi')
-
-    def test_charset_zang(self):
-        self.run_charset_validation('zang')
-
-    def run_charset_validation(self, charset):
+    @pytest.mark.parametrize("charset", test_charset_data)
```
pensivebrian, line 134: Using pytest, you no longer get integrated tests in vscode. Is there a reason why we're change this?

tests/test_globalization.py
ellbosch: That's a good point, I can use a tuple here for the test data.

tests/test_globalization.py
ellbosch: I don't believe that's true about Pytest anymore ([source](https://code.visualstudio.com/docs/python/testing)).

Pytest provides a lot of benefits over unittest, especially in readability of tests. In this particular instance we're using parametrized tests, enabling us to define one test method for multiple tests. This allowed us to remove redundant code.

pensivebrian: pytest doesn't work well with parameterized tests. Gene, the engineer before you, ported all pytests to be unittest tests due to this issue. I'd prefer we just leave them as is for now.

ellbosch: That's very interesting—truthfully I've already ported parametrized tests for 4 other test files (interactive, non_interactive, mssqlclient, multiline). Also looks like our original merge uses parametrized tests for ctes and parseutils.

I'll defer to your judgement if you still prefer to remove this, but in my experience it works great.

pensivebrian: As long as it easy to debug a particular language, without having to others, I'm okay with it I guess. What's the command line to run pytest with a specific parameter set?

ellbosch: Yes there's still the ability to log which language fails. I'm making a new commit to accommodate this.

I'm not positive on how to call a test from the command line with just a single parameter. [This StackOverflow post](https://stackoverflow.com/questions/44943470/parametrize-and-running-a-single-test-in-pytest) may show it's possible. I usually comment out the other parameters and call `pytest test_mod.py::TestClass::test_method`.

pensivebrian: So that's what I want to avoid - having to comment out tests to run a test.

ellbosch: I've validated you can test for a single parameter using the `-k` flag, followed by a string which matches the parameter to test.

ellbosch:
I just added another commit to improve reliability of our tests:
* Adds more cleanup that I should have added in the first place for non-interactive tests
* Creates test db for some tests that were running off of master db

ellbosch:
Another change today to improve DB create reliability and code quality:
* Tests in `test_interactive_mode.py` no longer use master DB.
* A new `test_db` fixture has been added to `mssqltestutils.py` to centralize redundant use of create and teardown of test DBs.

tests/mssqltestutils.py
```diff
@@ -137,3 +186,21 @@ def get_io_paths(test_file_suffix):
...
+
+
+class TestDB:
+    """
+    TestDB creates a test db for any tests inheriting this class, and cleans up resources when
+    the module is complete.
+    """
+    @staticmethod
+    @pytest.fixture(scope='module')
+    def test_db():
```
pensivebrian, line 198: creating an iterator for a single element to injecting the client seems overkill. Can we just have a setup method with creates the client?

tests/test_special.py
```diff
@@ -54,116 +61,97 @@ def tearDownClass(cls):
...
-                self.assertTrue(len(rows) == rows_expected_pattern_query)
-                self.assertTrue(len(col) == cols_expected_verbose)
-        finally:
-            shutdown(client)
 
+        for rows, col, _, _, _ in client.execute_query(command):
+            assert len(rows) >= min_rows_expected
+            assert len(col) == cols_expected
 
-if __name__ == u'__main__':
```
pensivebrian, line 168: These main's were useful to run files :(

ellbosch: I can put this back!

tests/mssqltestutils.py
ellbosch: I'm confused, I don't believe this is an iterator.

We're using Pytest [fixtures](https://docs.pytest.org/en/latest/fixture.html#scope-sharing-a-fixture-instance-across-tests-in-a-class-module-or-session) for tests which were already using this behavior, I was just trying to reduce redundant code.

ellbosch:
> I think the PR is becoming too complicated. I can tell what's stabilization and what is pytest migration. Can we just fix the create database issue, an address pytest some other time?

I can simplify this a bit—unfortunately, most of the code changes were about ensuring tests can run off of the master DB (which because a requirement now that we call `sys.dm_operation_status`).

However, I did refactor things a little in `test_special.py`, which I can save for another day 🙂

tests/mssqltestutils.py
```diff
@@ -90,28 +92,75 @@ def getTempPath(*args):
...
+    try:
         for _, _, status, _, is_create_error in client.execute_query(query_db_create):
-            if not is_create_error:
-                shutdown(client)
-                return test_db_name
-            # log warning to console
-            warnings.warn('Test DB create failed with error: {0}'.format(status))
-        count += 1
-    shutdown(client)
+            if is_create_error:
```
pensivebrian, line 112: This will prevent the retry logic from ever running.

tests/mssqltestutils.py
```diff
@@ -122,6 +118,10 @@ def create_test_db():
                     raise AssertionError("Database creation failed: retry logic for SQL " \
```
pensivebrian, line 118: Can we surface the status error message in the failure here as well?

ellbosch:
@pensivebrian I made a small tweak to our helper method to return an error message with the state response in the form of a tuple.

PR #448, opened 2020-04-08
Cast row_limit as integer

### LINKED ISSUE #444: Crash when displaying results if `--row-limit` is specified
# Description

By default, the `row-limit` is set to 1000 rows. According to the docs, you can override or even disable the prompt by specifying `--row-limit` when launching `mssql-cli`.

- `--row-limit 0`: disables the prompt
- `--row-limit <NUMBER>`: changes the default of `1000` to `<NUMBER>`

However, using this feature, a crash occurs when trying to display the rows:

```
Traceback (most recent call last):
  File "/usr/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/dylan/Projects/mssql/lib/python3.6/site-packages/mssqlcli/main.py", line 117, in <module>
    main()
  File "/home/dylan/Projects/mssql/lib/python3.6/site-packages/mssqlcli/main.py", line 110, in main
    run_cli_with(mssqlcli_options)
  File "/home/dylan/Projects/mssql/lib/python3.6/site-packages/mssqlcli/main.py", line 55, in run_cli_with
    mssqlcli.run()
  File "/home/dylan/Projects/mssql/lib/python3.6/site-packages/mssqlcli/mssql_cli.py", line 448, in run
    self.execute_query(text)
  File "/home/dylan/Projects/mssql/lib/python3.6/site-packages/mssqlcli/mssql_cli.py", line 375, in execute_query
    output = self._execute_interactive_command(text)
  File "/home/dylan/Projects/mssql/lib/python3.6/site-packages/mssqlcli/mssql_cli.py", line 331, in _execute_interactive_command
    output, query = self._evaluate_command(text)
  File "/home/dylan/Projects/mssql/lib/python3.6/site-packages/mssqlcli/mssql_cli.py", line 559, in _evaluate_command
    if self._should_show_limit_prompt(status, rows):
  File "/home/dylan/Projects/mssql/lib/python3.6/site-packages/mssqlcli/mssql_cli.py", line 527, in _should_show_limit_prompt
    return self.interactive_mode and self.row_limit > 0 and len(rows) > self.row_limit
TypeError: '>' not supported between instances of 'str' and 'int'
Exception ignored in: <bound method MssqlCli.__del__ of <mssqlcli.mssql_cli.MssqlCli object at 0x7f4ca0023518>>
Traceback (most recent call last):
  File "/home/dylan/Projects/mssql/lib/python3.6/site-packages/mssqlcli/mssql_cli.py", line 217, in __del__
  File "/home/dylan/Projects/mssql/lib/python3.6/site-packages/mssqlcli/sqltoolsclient.py", line 91, in shutdown
  File "/home/dylan/Projects/mssql/lib/python3.6/site-packages/mssqlcli/jsonrpc/jsonrpcclient.py", line 183, in shutdown
  File "/usr/lib/python3.6/logging/__init__.py", line 1308, in info
  File "/usr/lib/python3.6/logging/__init__.py", line 1444, in _log
  File "/usr/lib/python3.6/logging/__init__.py", line 1454, in handle
  File "/usr/lib/python3.6/logging/__init__.py", line 1516, in callHandlers
  File "/usr/lib/python3.6/logging/__init__.py", line 865, in handle
  File "/usr/lib/python3.6/logging/__init__.py", line 1071, in emit
  File "/usr/lib/python3.6/logging/__init__.py", line 1061, in _open
NameError: name 'open' is not defined
```

# Environment

**OS**: Ubuntu 18.04 LTS
**Python version**: Python 3.6.9
**Server version**: 15.0.4023.6 Developer Edition 64 bit

ellbosch:
Fixes #444.

The `--row-limit` arg was breaking mssql-cli because our code attempted to use a number operator on a string. The fix requires casting the parsed option as an integer.

Notes:
* `row_limit` is validated in the `MssqlCli` constructor, which uses a new `_set_row_limit` method to validate a positive integer.
* `_set_row_limit` calls `sys.exit(1)` if an invalid value is provided.
* As a consequence to the above, we needed to catch an `AttributeException` in the `__del__` method because shutdown is called on an uninstantiated `sqltoolsclient`.
* Testing included to hopefully prevent issues like this in the future.

### DIFF
--- a/mssqlcli/mssql_cli.py
+++ b/mssqlcli/mssql_cli.py
@@ -330,12 +330,15 @@ class MssqlCli(object):
         try:
             row_limit_int = int(row_limit)
         except ValueError:
-            click.secho(u'Error: --row-limit contains an invalid value as input. ' \
-                        u'Please use a positive integer.', fg='red')
+            click.secho(u'Error: row-limit has been set to an invalid value.\nPlease ' \
+                        u'specify a positive integer using the --row-limit command-line ' \
+                        u'argument or by setting row_limit in the config file.', fg='red')
             sys.exit(1)
         else:
             if row_limit_int < 0:
-                click.secho(u'Error: --row-limit cannot be negative.', fg='red')
+                click.secho(u'Error: row-limit cannot be negative.\nPlease ' \
+                            u'specify a positive integer using the --row-limit command-line ' \
+                            u'argument or by setting row_limit in the config file.', fg='red')
                 sys.exit(1)
             return row_limit_int
 
--- a/mssqlcli/mssql_cli.py
+++ b/mssqlcli/mssql_cli.py
@@ -167,8 +167,7 @@ class MssqlCli(object):
             self.multiline_mode = c['main'].get('multi_line_mode', 'tsql')
             self.vi_mode = c['main'].as_bool('vi')
             self.prompt_format = options.prompt or c['main'].get('prompt', self.default_prompt)
-            self.row_limit = \
-                self._set_row_limit(options.row_limit or c['main'].as_int('row_limit'))
+            self.row_limit = options.row_limit
             self.min_num_menu_lines = c['main'].as_int('min_num_menu_lines')
             self.multiline_continuation_char = c['main']['multiline_continuation_char']
             self.syntax_style = c['main']['syntax_style']
@@ -209,14 +208,9 @@ class MssqlCli(object):
                              "specified.")
 
     def __del__(self):
-        """ Shut down sqltoolsservice if defined. """
-        try:
-            # try/except block needed if shutdown is invoked before sqltoolsclient is defined.
-            # example: invalid row-limit argument is passed.
-            if self.sqltoolsclient:
-                self.sqltoolsclient.shutdown()            
-        except AttributeError:
-            pass
+        # Shut-down sqltoolsservice
+        if self.sqltoolsclient:
+            self.sqltoolsclient.shutdown()
 
     # TODO: possibly use at a later date for expanded output file functionality
     # def write_to_file(self, pattern, **_):
@@ -320,27 +314,6 @@ class MssqlCli(object):
 
             editor_command = special.editor_command(text)
         return text
-
-    @staticmethod
-    def _set_row_limit(row_limit):
-        """
-        Validates row_limit option has valid integer
-        """
-
-        try:
-            row_limit_int = int(row_limit)
-        except ValueError:
-            click.secho(u'Error: row-limit has been set to an invalid value.\nPlease ' \
-                        u'specify a positive integer using the --row-limit command-line ' \
-                        u'argument or by setting row_limit in the config file.', fg='red')
-            sys.exit(1)
-        else:
-            if row_limit_int < 0:
-                click.secho(u'Error: row-limit cannot be negative.\nPlease ' \
-                            u'specify a positive integer using the --row-limit command-line ' \
-                            u'argument or by setting row_limit in the config file.', fg='red')
-                sys.exit(1)
-            return row_limit_int
 
     def _execute_interactive_command(self, text):
         """ Runs commands in the interactive CLI mode. """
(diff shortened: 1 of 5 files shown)


mssqlcli/mssql_cli.py
```diff
@@ -167,11 +167,8 @@ def __init__(self, options):
             self.multiline_mode = c['main'].get('multi_line_mode', 'tsql')
             self.vi_mode = c['main'].as_bool('vi')
             self.prompt_format = options.prompt or c['main'].get('prompt', self.default_prompt)
-            if options.row_limit is not None:
-                self.row_limit = options.row_limit
-            else:
-                self.row_limit = c['main'].as_int('row_limit')
-
+            self.row_limit = \
```
pensivebrian, line 170: Is this the right layer/place to make the fix? This seems like an option parsing concern and perhaps handled by argparse?

mssqlcli/mssql_cli.py
```diff
@@ -212,9 +209,14 @@ def __init__(self, options):
...
 
     def __del__(self):
-        # Shut-down sqltoolsservice
-        if self.sqltoolsclient:
-            self.sqltoolsclient.shutdown()
+        """ Shut down sqltoolsservice if defined. """
+        try:
+            # try/except block needed if shutdown is invoked before sqltoolsclient is defined.
+            # example: invalid row-limit argument is passed.
+            if self.sqltoolsclient:
```
pensivebrian, line 216: This makes me nervous. In what situations will we raise an AttributeError?

mssqlcli/mssql_cli.py
ellbosch: So we definitely can add a check in argparse. However, we may want to keep logic elsewhere because we also accept a value from our config if there's no value for the `--row-limit` arg. Where that *somewhere* should be is open for debate.

I've considered `main.py` and the current location in `mssqlcli.py`. I picked the current location due to easier testability, also it seems like good design to have the constructor handle this logic.

mssqlcli/mssql_cli.py
ellbosch: `AttributeError` is raised f the `MssqlCli` constructor exits due to invalid parameters. In this case, say we input a series of characters instead of an integer for `--row-limit`—we exit during class instantiation, and before `sqltoolsclient` gets instantiated. Therefore the class doesn't possess the `sqltoolsclient` attribute, thus raising an `AttributeError`.

mssqlcli/mssql_cli.py
pensivebrian: Let's fix this to args parse for now - seems like the right place. 

Looks like a there are a number of issues reading values from our config so let's fix those in another PR at a later date.

ellbosch: I take back my prior statement—I was able to move logic for checking the config value right in argparse. It seems to work well, once tests pass I'll update the PR.

mssqlcli/mssqlclioptionsparser.py
```diff
@@ -187,3 +191,23 @@ def create_parser():
     )
 
     return args_parser
+
+def check_positive_int(row_limit):
```
pensivebrian, line 195: Can you add 'row_limit' to the method name? This way, it stands out from being a generic 'check_positive_int' method. Ideally, we have a generic method where check_row_limit calls check_positive_int('row_limit' value), and check positive can be resused.

ellbosch:
I'm fairly certain recent failures are due to PR validation build pipelines using the master DB, which tests in this branch don't support. Best we wait until #441 is approved.

pensivebrian:
Can you clarify what you mean by master db? master db should be read only, and tests should use their own database.

ellbosch:
> Can you clarify what you mean by master db? master db should be read only, and tests should use their own database.

Correct—with our stabilize DB PR, tests now generate a test DB if performing operations off of a database. Some tests were still performing commands off of the permanent DB we created, but that will all change once we merge #441.

In anticipation of this change, I set up or PR validation pipelines to no longer use the permanent DB, so we could properly test #441. This impacts other PRs (like this one) so it's best we wait to merge until #441 is complete.

Issue #457, opened 2020-04-20
Logs are too verbose


ellbosch:
Logs jump up in size > 5MB per session. They're too verbose, and some investigation is required to figure out why this is.

PR #458, opened 2020-04-20
Created buffer size for logs

### LINKED ISSUE #353: Logging files are massive
My machine's log files are in the order of 100's of MBs. We need a solution to limit the size of these logging files.

It appears that jsonrpc logging may be largely responsible here.

ellbosch:
Closes #353.

This caps logs at 100 MB and adds one backup file to extend our buffer, per recommendation in Python's docs.

I think part of the problem with our logging is that it's incredible verbose—the file size jumps up a lot with limited use. I've created another issue to focus on decreasing the verboseness of our logs (#457).

### DIFF
--- a/mssqlcli/mssql_cli.py
+++ b/mssqlcli/mssql_cli.py
@@ -2,6 +2,7 @@ import datetime as dt
 import functools
 import itertools
 import logging
+from logging.handlers import RotatingFileHandler
 import os
 import sys
 import threading
@@ -243,7 +244,9 @@ class MssqlCli(object):
         if log_level.upper() == 'NONE':
             handler = logging.NullHandler()
         else:
-            handler = logging.FileHandler(os.path.expanduser(log_file), encoding='utf-8')
+            # creates a log buffer with max size of 100 MB and one backup file
+            handler = RotatingFileHandler(os.path.expanduser(log_file), encoding='utf-8',
+                                          maxBytes=10**8, backupCount=1)
 
         level_map = {'CRITICAL': logging.CRITICAL,
                      'ERROR': logging.ERROR,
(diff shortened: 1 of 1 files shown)


mssqlcli/mssql_cli.py
```diff
@@ -243,7 +244,9 @@ def initialize_logging(self):
         if log_level.upper() == 'NONE':
             handler = logging.NullHandler()
         else:
-            handler = logging.FileHandler(os.path.expanduser(log_file), encoding='utf-8')
+            # creates a log buffer with max size of 100 MB and one backup file
+            handler = RotatingFileHandler(os.path.expanduser(log_file), encoding='utf-8',
```
pensivebrian, line 248: A 100MB file is still pretty large. Lets chunk the files into a size that someone can forward. How about 5 files with a max size of 20MB? Or 10 files 10MB?

ellbosch: Great idea, will change to 5 files/20 MB.

mssqlcli/mssql_cli.py
```diff
@@ -244,9 +244,9 @@ def initialize_logging(self):
         if log_level.upper() == 'NONE':
             handler = logging.NullHandler()
         else:
-            # creates a log buffer with max size of 100 MB and one backup file
+            # creates a log buffer with max size of 20 MB and 5 backup files
             handler = RotatingFileHandler(os.path.expanduser(log_file), encoding='utf-8',
-                                          maxBytes=10**8, backupCount=1)
+                                          maxBytes=2*(10**7), backupCount=5)
```
pensivebrian, line 249: Let's use 1024*1024*20 to specify 20MB, which is much more readable.

PR #463, opened 2020-04-24
Removed dev dependencies for deb and rpm

ellbosch:
Changes for both deb and rpm build scripts:
- Compiles two different Python exe's instead of just one
- Downloads dev dependencies in one version
- Installs the built mssql-cli wheel in the other version

I've validated that both new packages work and only come installed with dependencies in setup.py.

### DIFF
--- a/build_scripts/rpm/mssql-cli.spec
+++ b/build_scripts/rpm/mssql-cli.spec
@@ -8,17 +8,17 @@
   %define dist .el7
 %endif
 
-%define name                   mssql-cli
-%define release                1%{?dist}
-%define time_stamp             %(date +%y%m%d%H%M)
-%define base_version           1.0.0
-%define python_dir             %{_builddir}/python_env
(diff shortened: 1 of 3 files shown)


build_scripts/rpm/mssql-cli.spec
```diff
@@ -48,10 +50,14 @@ Requires:       libunwind, libicu, less
...
 # Clean previous build directory.
 rm -rf %{_builddir}/*
+rm -rf %{python_build_archive}
+
 # Download, Extract Python3
+mkdir %{python_build_archive}
 python_archive=$(mktemp)
 wget https://www.python.org/ftp/python/3.6.1/Python-3.6.1.tgz -qO $python_archive
 tar -xvzf $python_archive -C %{_builddir}
+tar -xvzf $python_archive -C %{python_build_archive}
```
ellbosch, line 60: change to `python_tmp_build`

PR #467, opened 2020-04-29
Added UTF-8 encoding on Python 2

ellbosch:
Fixes bug that breaks interactive queries when international strings are used in Python 2.


ellbosch:
Closing for different proposed fix

PR #505, opened 2021-01-27
Ubuntu 20.04 support

pensivebrian:
Add dependency for libicu66

### DIFF
--- a/build_scripts/debian/dir_creator.sh
+++ b/build_scripts/debian/dir_creator.sh
@@ -54,7 +54,7 @@ Homepage: https://github.com/dbcli/mssql-cli
 
 Package: mssql-cli
 Architecture: all
-Depends: libunwind8, libicu52 | libicu55 | libicu57 | libicu60, libffi-dev, less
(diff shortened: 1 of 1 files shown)


llali (approved):
:shipit:

pensivebrian:
Testing out the mssql-cli .deb package, I'm encountering errors:

ImportError: libffi.so.6: cannot open shared object file: No such file or directory

I think this is happening because we're compiling and redist-ing python on ubuntu 16, which uses the .6 version, but ubuntu 20 needs the .7 version. I'll confirm this, but if this is the case, we may need to starting shipping a .deb package per ubuntu version.

pensivebrian:
Building the .deb image with python compiled on Ubuntu 20.04 fixes the libffi.so.6 issue. So next step: how to manage shipping multiple .dep package versions which need to be output by the build.

danbrad:
Are there any updates on this? Ubuntu 20.04 has been out a while now and is an LTS release so it would be nice to be able to upgrade to it.

